PR #51, opened 2017-09-26
Update README.md

lukabratos:
### What?

Update link to the docs.

----
CC @pusher/mobile


hamchapman:
I think the links, given we're in the context of the Swift repo, should go to `https://docs.pusher.com/chatkit/reference/swift`

PR #59, opened 2018-01-06
🐛 Fixed bug where the user_id was not retrieved

neoighodaro:
🐛 Fixed bug where the user_id was not retrieved

### What?
The `user_id` key is not in the root object. Therefore `user_id` will be `nil` as it was not being retrieved properly thus causing a crash.

...

----
CC @pusher/mobile


hamchapman:
Thanks!

PR #62, opened 2018-02-12
Cursors

hamchapman:
### What?

- Add cursors support
- Add connection coordinator; connection completion handler(s) are now only called when the following have completed:
   * User subscription has been established (or errored)
   * Presence subscription has been established (or errored)
   * Initial cursors fetch has completed (or errored)
   * Initial users fetch has completed (or errored)

### Why?

To add support for cursors and make connection behaviour more useful

----
CC @pusher/mobile

### DIFF
--- a/Sources/PCRoom.swift
+++ b/Sources/PCRoom.swift
@@ -11,6 +11,18 @@ public final class PCRoom {
     public internal(set) var deletedAt: String?
 
     public internal(set) var subscription: PCRoomSubscription?
+    public internal(set) var cursorSubscription: PCCursorSubscription?
+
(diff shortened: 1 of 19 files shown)


Sources/PCRoom.swift
```diff
@@ -11,6 +11,18 @@ public final class PCRoom {
     public internal(set) var deletedAt: String?
 
     public internal(set) var subscription: PCRoomSubscription?
+    public internal(set) var cursorSubscription: PCCursorSubscription?
+
+    public internal(set) var currentUserCursor: PCBasicCursorState?
```
hot-leaf-juice: 👌

PR #65, opened 2018-02-16
Provide SDK info object to Instance(s) created as part of ChatManager instantiation

hamchapman:
### What?

See title

### Why?

----

CC @hamchapman


hamchapman:
[Add Swift client library and version as request headers so that we know what SDKs our customers are using](https://trello.com/c/fESU4wco/603-add-swift-client-library-and-version-as-request-headers-so-that-we-know-what-sdks-our-customers-are-using)

PR #67, opened 2018-02-26
Use separate presence instance to communicate with chatkit_presence

hamchapman:
### What?

See title

### Why?

We want to use the new presence service

----

CC @hamchapman


hamchapman:
[Update Swift SDK to support new presence service route](https://trello.com/c/mUPxrl4z/674-update-swift-sdk-to-support-new-presence-service-route)

PR #69, opened 2018-03-19
Cursor revamp and associated changes

hamchapman:
### What?

* `subscribeToRoom` attempts to join the room if necessary.
* Establish user cursor subscription on connection.
* `PCRoomSubscription` -> `PCMessageSubscription`.
* `PCRoomSubscription` now holds reference to a `PCMessageSubscription` and a `PCCursorSubscription`.

### Why?

More closely match the JS SDK

----

CC @hamchapman

### DIFF
--- a/Sources/PCBasicCurrentUser.swift
+++ b/Sources/PCBasicCurrentUser.swift
@@ -1,0 +1,170 @@
+import Foundation
+import PusherPlatform
+
+public final class PCBasicCurrentUser {
+    public let id: String
+    public let pathFriendlyId: String
+
+    let userStore: PCGlobalUserStore
+    let roomStore: PCRoomStore
+    let cursorStore: PCCursorStore
+
+    public internal(set) var userSubscription: PCUserSubscription?
+    public internal(set) var presenceSubscription: PCPresenceSubscription?
+    public internal(set) var cursorSubscription: PCCursorSubscription?
+
+    let instance: Instance
+    let filesInstance: Instance
+    let cursorsInstance: Instance
+    let presenceInstance: Instance
+
+    let connectionCoordinator: PCConnectionCoordinator
+
+    public init(
+        id: String,
+        pathFriendlyId: String,
+        instance: Instance,
+        filesInstance: Instance,
+        cursorsInstance: Instance,
+        presenceInstance: Instance,
+        connectionCoordinator: PCConnectionCoordinator
+    ) {
+        self.id = id
+        self.pathFriendlyId = pathFriendlyId
+        self.instance = instance
+        self.filesInstance = filesInstance
+        self.cursorsInstance = cursorsInstance
+        self.presenceInstance = presenceInstance
+        self.connectionCoordinator = connectionCoordinator
+
+        let rooms = PCSynchronizedArray<PCRoom>()
+        self.userStore = PCGlobalUserStore(instance: instance)
+        self.roomStore = PCRoomStore(rooms: rooms, instance: instance)
+        self.cursorStore = PCCursorStore(
+            instance: instance,
+            roomStore: roomStore,
+            userStore: userStore
+        )
+    }
+
+    func establishUserSubscription(
+        delegate: PCChatManagerDelegate,
+        initialStateHandler: @escaping ((roomsPayload: [[String: Any]], currentUserPayload: [String: Any])) -> Void
+    ) {
+        let path = "/users"
+        let subscribeRequest = PPRequestOptions(method: HTTPMethod.SUBSCRIBE.rawValue, path: path)
+
+        var resumableSub = PPResumableSubscription(
+            instance: self.instance,
+            requestOptions: subscribeRequest
+        )
+
+        self.userSubscription = PCUserSubscription(
+            instance: self.instance,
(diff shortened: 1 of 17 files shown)


hamchapman:
[Make subscribeToRoom in Swift SDK behave the same as in the new JS SDK (i.e. join if user not already joined)](https://trello.com/c/mXtNdIf4/741-make-subscribetoroom-in-swift-sdk-behave-the-same-as-in-the-new-js-sdk-ie-join-if-user-not-already-joined)

Sources/PCMessageSubscription.swift
```diff
@@ -0,0 +1,73 @@
...
+
+        guard eventTypeName == expectedEventTypeName else {
+            self.logger.log("Expected event type name to be \(expectedEventTypeName) but got \(eventTypeName)", logLevel: .debug)
+            return
+        }
+
+        guard let messagePayload = json["data"] as? [String: Any] else {
+            self.logger.log("Missing data for room subscription event: \(json)", logLevel: .debug)
+            return
+        }
```
pablisco: The returns on these methods seem to lose potential errors. Should these be reported to the presentation layer?

Sources/PCUserSubscription.swift
```diff
@@ -636,6 +488,7 @@ extension PCUserSubscription {
             return
         }
 
+        // TODO: Why not weak self here?
```
pablisco: Should these `TODO` have a card?

Sources/PCRoomSubscription.swift
```diff
@@ -2,67 +2,26 @@ import Foundation
...
-    let basicMessageEnricher: PCBasicMessageEnricher
+    let messageSubscription: PCMessageSubscription
+    let cursorSubscription: PCCursorSubscription
+
+    // TODO: Maybe the RoomSubscription should just store the delegate?
+    public var delegate: PCRoomDelegate? {
+        get {
+            return messageSubscription.delegate
+        }
+    }
```
pablisco: Wouldn't it be better to provide the delegate in the constructor instead of having a `var`?

Sources/PCMessageSubscription.swift
pablisco: This also applies to other logger side effects in here

Sources/PCBasicCurrentUser.swift
```diff
@@ -0,0 +1,170 @@
...
+            cursorsInstance: self.cursorsInstance,
+            presenceInstance: self.presenceInstance,
+            resumableSubscription: resumableSub,
+            userStore: self.userStore,
+            delegate: delegate,
+            userId: id,
+            pathFriendlyUserId: pathFriendlyId,
+            connectionCoordinator: connectionCoordinator,
+            initialStateHandler: initialStateHandler
+        )
```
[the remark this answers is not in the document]
hamchapman: Yeah definitely, there's a lot of refactoring that I want to get to eventually to tidy this all up. Functions taking 10+ initialisers is...suboptimal 😛

Sources/PCBasicCursorEnricher.swift
```diff
@@ -3,40 +3,56 @@ import PusherPlatform
...
         self.userStore = userStore
-        self.room = room
+        self.roomStore = roomStore
         self.logger = logger
     }
 
     func enrich(_ basicCursor: PCBasicCursor, completionHandler: @escaping (PCCursor?, Error?) -> Void) {
-        self.userStore.user(id: basicCursor.userId) { [weak self] user, err in
+        self.userStore.user(id: basicCursor.userId) { [weak self] user, userErr in
             guard let strongSelf = self else {
```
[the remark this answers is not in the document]
hamchapman: Could do, but then the `userStore.user` call would have to support that as an extra bit of functionality. That's not to say that that's necessarily a bad thing. The way it is currently just means that all we need is a `[weak self]` as a capture group and the `guard` statement just to be extra sure that `self` hasn't been completely dereferenced.

Sources/PCRoomSubscription.swift
hamchapman: Potentially - I hadn't made my mind up yet, hence the TODO: https://github.com/pusher/chatkit-swift/pull/69/files#diff-6ae7c192bfbc593d33d9f6769b3f8ea3R8 

I was thinking of making the delegate a non-optional property of the two subscriptions as well, but hadn't fully thought through whether or not that was too big of a breaking change. 

I'll take a proper look and think about it 👍

Sources/PCUserSubscription.swift
hamchapman: Maybe - they're not high priority and more gentle reminders for me (or someone else) to think about when they're in the code next. I just spotted them as a passing thing

Sources/PCMessageSubscription.swift
hamchapman: It's an interesting one. I come down on the no side because the expectation is that events received from the server should always have all of this information; the error case should be extremely, extremely rare. If they don't then that's an indication that something has gone wrong in an unexpected way. If that's the case, then we're putting the burden of determining whether or not to communicate something with the end user, which is not going to be obvious as it's not something that the user has done that would have led to an error of some kind.

PR #70, opened 2018-03-20
Queue token fetch requests if one is already underway

hamchapman:
### What?

Queue token fetch requests if one is already underway

### Why?

Fewer requests made to token endpoint

----

CC @hamchapman


hamchapman:
[Don't make more token requests while the first one is still in progress](https://trello.com/c/MXFXQgVx/771-dont-make-more-token-requests-while-the-first-one-is-still-in-progress)

PR #71, opened 2018-04-06
Fix sending messages when an attachment is not set

steve228uk:
### What?

Currently messages were not sending if an attachment was not set. This fixes that.

----

CC @hamchapman

### DIFF
--- a/Sources/PCCurrentUser.swift
+++ b/Sources/PCCurrentUser.swift
@@ -691,33 +691,36 @@ public final class PCCurrentUser {
         text: String? = nil,
         attachmentType: PCAttachmentType? = nil,
         completionHandler: @escaping (Int?, Error?) -> Void
-    ) {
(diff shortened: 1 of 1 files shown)


hamchapman:
Thanks for this! Looks like quite a large oversight on my part. Sorry about that!

I'll make sure that I add some tests for this on Monday so it can't happen in future.

steve228uk:
@hamchapman Did you manage to look into getting some tests written for this? It would be great to have it merged if possible 😊

hamchapman:
@steve228uk I did but got caught up in trying to get stubbing of a HTTP/2 streamed request working!

I think I'll merge this for now and follow up with the tests later once I've got the stubbing working properly. Thanks again!

steve228uk:
@hamchapman You're a hero 😄 Thanks so much!

Issue #76, opened 2018-04-24
Memory Leak: PCRoomSubscription retaining strong reference to PCRoomDelegate


steve228uk:
### What?

When calling the following method on `PCCurrentUser` a delegate is passed to the method. However the declaration for the delegate on `PCRoomSubscription` is not marked as weak. This causing the `deinit` on the class assigned as the delegate to never be called.

```swift
func subscribeToRoom(room: PCRoom, roomDelegate: PCRoomDelegate, messageLimit: Int = 20)
```

The `PCRoomDelegate` should be marked as compliance with `NSObjectProtocol` to allow it to be defined a weak reference to avoid this happening.

As the `subscription` property is only able to be set internally by ChatKit it's not possible to sever that connection. The `end()` method on `PCRoomSubscription` should also likely set the delegate to nil to avoid that reference being kept.

### Environment

* **PusherChatkit version:** 0.7.2
* **Xcode version:** 9.4
* **Swift version:** 4.1
* **Platform(s) running PusherChatkit:** iOS

----

CC @hamchapman

steve228uk:
It seems that this may not be the only aspect causing the memory leak. I setup the weak link in my fork and it's still retaining a strong reference. Could this be because the `PCRoom` is never removed from memory because the `PCCurrent` user keeps a reference to it and all of its subscriptions? Is there a way to manually unsubscribe or keep a weak link to a room there?

steve228uk:
Here's the memory graph for the retained instance of the delegate. `PCRoomSubscription` is no longer holding a reference to it in my fork but `PPSubscriptionDelegate` is. I think this is caused by the non-weak reference to the `PCRoomDelegate` on `PCMessageSubscription`.

<img width="1070" alt="screen shot 2018-04-24 at 12 43 35" src="https://user-images.githubusercontent.com/1169297/39184891-210f0716-47bd-11e8-8bdd-455284760bd7.png">

hamchapman:
Fixed in #77

PR #77, opened 2018-04-24
Fix memory leak with the PCRoomDelegate

steve228uk:
### What?

This PR fixes the memory leak caused by strong references to `PCRoomDelegate`, `PCMessageSubscription` and `PCCursorSubscription` being retained.

More details can be found in #76.

----

CC @hamchapman

### DIFF
--- a/Sources/ChatManager.swift
+++ b/Sources/ChatManager.swift
@@ -271,7 +271,7 @@ import PusherPlatform
                                 logLevel: .debug
                             )
                             if roomUsersProgressCounter.incrementFailedAndCheckIfFinished() {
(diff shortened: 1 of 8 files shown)


hamchapman:
This is great - thanks @steve228uk 

I think there's probably still a few other places that need to be considered in terms of whether or not a strong or weak reference should be being held (e.g. the various things that keep hold of `Instance` objects) but this is a great start, especially if it fixes your issues!

steve228uk:
Thanks @hamchapman 😃 If I spot any others I'll submit another PR.

PR #78, opened 2018-04-27
Add debouncing for setting read cursors

hamchapman:
### What?

Add debouncing for setting read cursors

### Why?

So the service doesn't get hammered by loads of read cursor requests

----

CC @hamchapman

### DIFF
--- a/Sources/PCReadCursorDebouncerManager.swift
+++ b/Sources/PCReadCursorDebouncerManager.swift
@@ -1,0 +1,98 @@
+import Foundation
+
+class PCReadCursorDebouncerManager {
+    private var roomIdsToDebouncers: [Int: PCReadCursorDebouncer] = [:]
+    private weak var currentUser: PCCurrentUser?
+
+    init(currentUser: PCCurrentUser) {
+        self.currentUser = currentUser
+    }
+
+    func set(cursorPosition position: Int, inRoomId roomId: Int, completionHandler: @escaping PCErrorCompletionHandler) {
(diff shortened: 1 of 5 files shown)


Sources/PCReadCursorDebouncerManager.swift
```diff
@@ -0,0 +1,81 @@
...
+        self.interval = Double(intervalMilliseconds  / 1000)
+    }
+
+    deinit {
+        self.timer?.invalidate()
+        self.timer = nil
+    }
+
+    func set(position: Int, completionHandler: @escaping PCErrorCompletionHandler) {
+        sendReadCursorPayload = (position: position, completionHandler: completionHandler)
```
callum-oakley: I think the `max` permission should be respected (rather than just the last one). If you're calling `setReadCursor` asynchronously for a load of positions, you want the greatest one to be respected.

Sources/PCReadCursorDebouncerManager.swift
```diff
@@ -0,0 +1,81 @@
...
+
+        guard let payload = self.sendReadCursorPayload else {
+            print("sendReadCursorPayload is nil so can't set read cursor")
+            return
+        }
+
+        currentUser.sendReadCursor(
+            position: payload.position,
+            roomId: roomId,
+            completionHandler: payload.completionHandler
```
callum-oakley: Do we only want the last completion handler to be called? Or do we want every completion handler to be called? I went with the latter in the JS because then the external interface is the same as it was... not a big deal though.

PR #83, opened 2018-07-12
Cursors tests

callum-oakley:
### What?

Tests the existing behaviour. There's no hook for receiving changes to your own read cursors if you're not subscribed to the room, I'll stick a card in triage.

----

CC @hamchapman

### DIFF
--- a/Tests/CursorTests.swift
+++ b/Tests/CursorTests.swift
@@ -2,65 +2,99 @@ import XCTest
 import PusherPlatform
 @testable import PusherChatkit
 
-func testUser(userId: String) -> ChatManager {
-    return ChatManager(
-        instanceLocator: testInstanceLocator,
-        tokenProvider: PCTokenProvider(url: testInstanceTokenProviderURL),
-        userId: userId,
-        logger: TestLogger()
-    )
-}
+var alice: PCCurrentUser?
+var bob: PCCurrentUser?
+var roomId: Int?
+var doneSetUp = false
 
 class CursorTests: XCTestCase {
-    var aliceManager: ChatManager?
-    var bobManager: ChatManager?
-    var room: PCRoom?
-    
-    // Currently these tests only pass if the users "alice" and "bob" already exist.
     override func setUp() {
-        let ex = expectation(description: "Room is created")
-        if aliceManager == nil {
-            aliceManager = testUser(userId: "alice")
-        }
-        if bobManager == nil {
-            aliceManager = testUser(userId: "bob")
-        }
-        if room == nil {
-            aliceManager?.connect(delegate: TestingChatManagerDelegate()) { alice, error in
-                XCTAssertNil(error)
-                
-                alice?.createRoom(name: "Alice's room") { r, error in
-                    XCTAssertNil(error)
-                    
-                    self.room = r
-                    
-                    ex.fulfill()
-                }
-            }
-        } else {
-            ex.fulfill()
-        }
         super.setUp()
-        waitForExpectations(timeout: 5)
+
+        if !doneSetUp {
+            alice = user(id: "alice")
+            bob = user(id: "bob")
+            roomId = createRoom(user: user(id: "alice"), roomName: "mushroom", addUserIds: ["bob"]).id
+            doneSetUp = true
+        }
     }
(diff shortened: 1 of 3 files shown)


Tests/CursorTests.swift
```diff
@@ -0,0 +1,194 @@
...
+
+    // TODO hook for setting own read cursor? (currently unsupported by the looks of it)
+
+    func testGetOwnReadCursor() {
+        let ex = expectation(description: "got own read cursor")
+
+        alice?.setReadCursor(position: 42, roomId: roomId!) { error in
+            XCTAssertNil(error)
+
+            sleep(1) // give the read cursor a chance to propagate down the connection
```
hamchapman: Is this required if we're in the completion handler closure? Presumably it should already be set and so we don't need to sleep?

Tests/CursorTests.swift
```diff
@@ -0,0 +1,194 @@
...
+                    XCTFail()
+                }
+            }
+        }
+
+        waitForExpectations(timeout: 5)
+    }
+
+    func testGetAnotherUsersReadCursor() {
+        let ex = expectation(description: "got another users read cursor fails")
```
hamchapman: Probs want to change the description as this is the success case, although not really that important.

Tests/CursorTests.swift
```diff
@@ -0,0 +1,194 @@
...
+            room: alice!.rooms.first(where: { $0.id == roomId! })!,
+            roomDelegate: aliceRoomDelegate
+        )
+
+        sleep(1)
+
+        bob?.setReadCursor(position: 42, roomId: roomId!) { error in
+            XCTAssertNil(error)
+
+            sleep(1) // give the read cursor a chance to propagate down the connection
```
hamchapman: Again, not sure we need this. Maybe I'm missing something though...

Tests/CursorTests.swift
```diff
@@ -0,0 +1,194 @@
...
+    func testGetAnotherUsersReadCursor() {
+        let ex = expectation(description: "got another users read cursor fails")
+
+        let aliceRoomDelegate = AliceRoomDelegate()
+        alice!.subscribeToRoom(
+            room: alice!.rooms.first(where: { $0.id == roomId! })!,
+            roomDelegate: aliceRoomDelegate
+        )
+
+        sleep(1)
```
hamchapman: Perhaps add a `TODO` to remove this once we've made `subscribetoRoom` return a promise or something like that?

Tests/CursorTests.swift
```diff
@@ -0,0 +1,194 @@
...
+            user = u
+            ex.fulfill()
+        }
+
+        waitForExpectations(timeout: 5)
+        return user!
+    }
+
+    func createRoom(user: PCCurrentUser, roomName: String, addUserIds: [String] = []) -> PCRoom {
+        var room: PCRoom?
```
hamchapman: This is just a comment on usage of Swift more than anything but you could define room as `PCRoom!` here and then when you return you would just `return room` (i.e. no need for the `room!` in the return)

Tests/CursorTests.swift
callum-oakley: If you remove it the tests fail some of the time, I think that what's happening is the request has succeeded, but we haven't received the cursor back on the subscription yet, and _thats_ when it gets updated internally. In fact we would always need at least a small timeout because if you're setting a cursor as one user and reading it as another there's always going to be a small gap between the request suceeding and the state updating.

Tests/CursorTests.swift
callum-oakley: oops! good catch

Tests/CursorTests.swift
callum-oakley: noooooice. This is exactly what I actually wanted.

Tests/CursorTests.swift
callum-oakley: @hamchapman this has been marked out of date because of an unrelated change, pinging you so it doesn't get lost

hamchapman: Ah good point - guess we'll have to live with it for now then

PR #84, opened 2018-07-12
Get the stubbed test working as a basis for writing more stubbed tests

hamchapman:
### What?

See title

### Why?

----

CC @hamchapman

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ fastlane/report.xml
 fastlane/screenshots
 
 .swift-version
+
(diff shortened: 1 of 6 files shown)


callum-oakley:
I think we can cut down the stubs a lot by using the provided matchers and builders. e.g. for the first one:

```swift
stub(uri("\(tokenEndpoint)?user_id=\(userID)"), json([
    "access_token": "a.good.token",
    "expires_in": 86400
]))
```

(and `jsonData` for the ones where we just provide the raw json string)

PR #91, opened 2018-07-23
Membership (and presence) tests

hamchapman:
### What?

Contains all of the membership tests and presence tests (which can be considered reviewed and 👍-ed).

### Why?

Tests are good

----

CC @hamchapman

### DIFF
--- a/Tests/RoomMembershipTests.swift
+++ b/Tests/RoomMembershipTests.swift
@@ -342,4 +342,37 @@ class RoomMembershipTests: XCTestCase {
 
         waitForExpectations(timeout: 10)
     }
+
+    // MARK: users property tests
+
+    func testUsersPropertyOfRoomIsProperlyPopulatedAfterSubscribingToRoom() {
+        let usersSetProperly = expectation(description: "users property is correct")
+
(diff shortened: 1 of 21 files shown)


Tests/RoomMembershipTests.swift
```diff
@@ -0,0 +1,471 @@
...
+            userJoinedRoomHookEx.fulfill()
+        }
+
+        let aliceCMDelegate = TestingChatManagerDelegate(userJoinedRoom: userJoinedRoom)
+
+        self.aliceChatManager.connect(delegate: aliceCMDelegate) { alice, err in
+            guard err == nil else {
+                XCTFail(err!.localizedDescription)
+                return
+            }
```
callum-oakley: is this better than `XCTAssertNil(err)`?

hamchapman: Ha, good point. Probably not - I'll use `XCTAssertNil`. Nice and concise, thanks!

hamchapman:
Added a test for the `users` property and replaced all of the `guard`s with `XCTAssertNil`s

Going to merge now

PR #92, opened 2018-07-25
subscription completion handler

callum-oakley:
I have a feeling this might not be very idiomatic, in which case suggestions of alternative approaches would be great!

----

CC @hamchapman

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - `sendMessage` now requires a value for the `text` parameter
 - All `PCMessage`s must have a `text` property
+- `subscribeToRoom` now has a (required) completion handler of type
+  `PCErrorCompletionHandler`
 
 ## [0.8.4](https://github.com/pusher/chatkit-swift/compare/0.8.3...0.8.4) - 2018-05-26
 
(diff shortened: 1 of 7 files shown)


hamchapman:
I think `DispatchGroup` is more idiomatic. I also made a homegrown thing to handle this sort of async stuff: https://github.com/pusher/chatkit-swift/blob/master/Sources/PCProgressCounter.swift

We should also get rid of that and replace its usage with `DispatchGroup`s, but that's a bigger job as it's used in quite a few places.

In theory your `steppedCompletionHandler` should work fine, but the `wait()` makes me a bit nervous. I think that will block whichever thread that the completion handler is running on, which could feasibly be the main thread ("UI thread") and we don't want to block that. I'm not 100% sure on that though.

Ultimately, if you're happy to give `DispatchGroup`s a go in this case, that'd be great, otherwise we can roll with this for now and make using `DispatchGroup`s a separate thing to tackle as a whole.

callum-oakley:
Ahh cool, I didn't see the progress counter. I'll try with dispatch groups.

callum-oakley:
@hamchapman tried using DispatchGroup

hamchapman:
LGTM with the `DispatchGroup` 👍 

Whoever gets to this first should rebase on master and check that everything still works.

The `MessageTests` (now on master) will likely need to have the same sleep-removal-and-update-with-completion-handler treatment as a result.

callum-oakley:
@hamchapman Looks like you rebased, can we merge this?

hamchapman:
Just waiting for the tests to pass... 🤞

PR #97, opened 2018-07-30
order fetch messages by increasing id

callum-oakley:
To match the Javascript SDK...

*But* I do see the logic in having it go the other way 🤔 When you're paginating it's certainly nicer to have the newest items first. Maybe we should change it in the js sdk instead? wdyt @hamchapman?


hamchapman:
I think this makes more sense. It will then match how messages are ordered when received as part of a room subscription being established, where we call `newMessage` with the oldest first.

lukejacksonn:
I never found the JS way particularly unintuitive or hard to work with so I'd say go with this way.

callum-oakley:
Cool cool. Thanks both. :)

callum-oakley:
Treating the above and the fact that mike put my face back on the card as implicit approval 😅

PR #99, opened 2018-08-08
Get users by ids v2

vivangkumar:
### What?

Get users by ids now takes query parameters in a more sane way with repeating query parameters that get interpreted as an array on the server side.

### Why?

because `,` could be a valid user id, which leads to all sorts of disgraces.

----

CC @hamchapman

### DIFF
--- a/Sources/PCGlobalUserStore.swift
+++ b/Sources/PCGlobalUserStore.swift
@@ -100,11 +100,9 @@ public final class PCGlobalUserStore {
             return
         }
 
-        let userIdsString = userIds.joined(separator: ",")
-
         let path = "/users_by_ids"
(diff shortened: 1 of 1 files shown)


vivangkumar:
[[3] Ensure parameters read from URLs are always correctly decoded (v2)](https://trello.com/c/oexc1Crw/1240-3-ensure-parameters-read-from-urls-are-always-correctly-decoded-v2)

hamchapman:
Pretty much all the tests failed because the CI is set up to make requests to us1 and that hasn't been updated to support the v2 version of getting users by IDs.

Change, code wise, LGTM to though, so feel free to merge into api-v2 👍

vivangkumar:
Ah, I was wondering that might be the case. Ty.

PR #100, opened 2018-08-13
Subscribe to all rooms on connection

hamchapman:
### What?

* Subscribe to all rooms upon connection and before returning the `PCCurrentUser`
* No more `PCRoomDelegate` because no more `subscribeToRoom` calls intended to be made directly (you can still join but then the idea is that the subscription will be established automatically)

Interesting bits:

* `subscribeToAllRooms` called one the initial user subscription completes successfully: https://github.com/pusher/chatkit-swift/pull/100/files#diff-cefb9d807d95970c3bd5f9fa7d60e804R188
* `preCompletionHooks` added to `PCRoomStore` so that once a new room is added to the room store (user joins room or is added to room, either explicitly or because of a room creation), all of the `preCompletionHooks` need to complete before the room is considered added to the store. I think there might be a better (read cleaner) abstraction here, but I couldn't really think of one while I was doing this (see the `addOrMerge` function for the meat of it): https://github.com/pusher/chatkit-swift/pull/100/files#diff-51e34dfe4d85b851d06954a5f6c6ddadR8
* `connect()` callsite is now where you can specify how many messages you want to receive as part of an initial room subscription: https://github.com/pusher/chatkit-swift/pull/100/files#diff-cefb9d807d95970c3bd5f9fa7d60e804R86

### Why?

Simplify the client

----

CC @hamchapman


hamchapman:
Closing in favour of #103

PR #103, opened 2018-08-17
Subscribe all rooms on connect

hamchapman:
### What?

* Subscribe to all rooms upon connection and before returning the `PCCurrentUser`
* No more `PCRoomDelegate` because no more `subscribeToRoom` calls intended to be made directly (you can still join but then the idea is that the subscription will be established automatically)

Interesting bits:

* `subscribeToAllRooms` called one the initial user subscription completes successfully: https://github.com/pusher/chatkit-swift/pull/100/files#diff-cefb9d807d95970c3bd5f9fa7d60e804R188
* `preCompletionHooks` added to `PCRoomStore` so that once a new room is added to the room store (user joins room or is added to room, either explicitly or because of a room creation), all of the `preCompletionHooks` need to complete before the room is considered added to the store. I think there might be a better (read cleaner) abstraction here, but I couldn't really think of one while I was doing this (see the `addOrMerge` function for the meat of it): https://github.com/pusher/chatkit-swift/pull/100/files#diff-51e34dfe4d85b851d06954a5f6c6ddadR8
* `connect()` callsite is now where you can specify how many messages you want to receive as part of an initial room subscription: https://github.com/pusher/chatkit-swift/pull/100/files#diff-cefb9d807d95970c3bd5f9fa7d60e804R86

### Why?

Simplify the client

----

CC @hamchapman


teminer:
Before merging this, we need to discuss product tradeoffs. Making this automatic without the option to disable it means some scenarios (like agencies who have multiple customers on one instance) will break their use case. We should discuss how to make this a bit more flexible

hamchapman:
This was only left here in case we decide we _do_ want to use it in future. I can close the PR and just leave the branch if that's cleaner

teminer:
It's fine to leave as-is... I just wanted to comment that if we do decide in the future, we need to discuss a bit more fully. :)

clay-mindoula:
Me again, I'm also interested in this one, very confused about who should be doing what I was. Just wondering on an ETA to schedule in some refactoring? thanks!

clay-mindoula:
Ah, oh well. Worked for my "Use Case" but I can live with what I have. thanks again for the prompt response.

Issue #107, opened 2018-10-25
createdAtDate on PCMessage causing crashes at times


steve228uk:
### What?

`createdAtDate` on PCMessage force unwraps a date created from a string. There seems to be times when this has caused a crash for our users.

I've been unable to replicate the issue but I have but have seen crash reports coming in.

### Environment

* **PusherChatkit version:** 0.8.1
* **Xcode version:** 10.0
* **Swift version:** 4.1
* **Platform(s) running PusherChatkit:** iOS

----

CC @hamchapman

hamchapman:
Do you have any more info on what the `createdAt` string value is for when the crash occurs?

steve228uk:
Unfortunately not, @hamchapman 😟 I've only got the stack from the iOS crash report. It's coming from the `newMessage(message: PCMessage)` delegate method. When attempting to insert that into a collection view and read the value from `createdAtDate` the getter throws an error.

It's not something that we've been able to replicate it ourselves but have had multiple users and multiple crash reports from multiple devices through.

lukabratos:
We might try and set the locale in the [DateFormatter](https://github.com/pusher/chatkit-swift/blob/master/Sources/PCMessage.swift#L12).

https://developer.apple.com/library/archive/qa/qa1480/_index.html

hamchapman:
What's the workaround you've been using @steve228uk ?

steve228uk:
@hamchapman I'm using a date formatter on the `createdAt` date but not force unwrapping. 

```swift
let dateFormatter = DateFormatter()
dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
let sentAt = dateFormatter.date(from: message.createdAt) ?? Date()
```

clay-mindoula:
i just got this also:
````
self?.chatRooms[roomIndex].timeAgo = message.createdAtDate
```
Happened just the once so far, it was in the Sym alas so no crash report.

hamchapman:
If you're able to, please try using the branch that forms #128 to see if that causes the crashes to go away

clay-mindoula:
i'll take a look early next week, many thanks

hamchapman:
The expected fix for this is now on master so if anyone has a chance to test that before I cut a new release then that would be much appreciated!

PR #115, opened 2018-12-04
deletedAt field on room

callum-oakley:
CC @hamchapman

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
 github "hamchapman/Mockingjay" "90156b01840e39549a451f97fc5b611ecf060d8f"
-github "krzyzanowskim/CryptoSwift" "0.13.0"
+github "krzyzanowskim/CryptoSwift" "0.13.1"
 github "pusher/beams-chatkit-swift" "1.2.2"
 github "pusher/pusher-platform-swift" "0.6.2"
--- a/Sources/PCRoom.swift
+++ b/Sources/PCRoom.swift
@@ -8,6 +8,7 @@ public final class PCRoom {
     public let createdByUserID: String
     public let createdAt: String
     public internal(set) var updatedAt: String
+    public internal(set) var deletedAt: String?
     public internal(set) var customData: [String: Any]?
 
     public internal(set) var subscription: PCRoomSubscription?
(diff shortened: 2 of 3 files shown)


Sources/PCRoom.swift
```diff
@@ -8,6 +8,7 @@ public final class PCRoom {
     public let createdByUserID: String
     public let createdAt: String
     public internal(set) var updatedAt: String
+    public internal(set) var deletedAt: String?
```
hamchapman: We also will want to add an accompanying `public var deletedAtDate: Date { return self.dateFormatter.date(from: self.deletedAt)! }` (to match `createdAt` and `updatedAt`)

Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
 github "hamchapman/Mockingjay" "90156b01840e39549a451f97fc5b611ecf060d8f"
-github "krzyzanowskim/CryptoSwift" "0.13.0"
+github "krzyzanowskim/CryptoSwift" "0.13.1"
```
hamchapman: If there's a change in the `Cartfile.resolved` we should reflect it in the `Podfile` too (i.e. bump the dep to 0.13.1 there too)

callum-oakley: Where? `PusherChatkit.podspec`? I can't find any other references to `CryptoSwift`

hamchapman: Ah yes, my bad. Forgot it was just a dependency for the tests. Ignore me

Sources/PCRoom.swift
```diff
@@ -35,6 +36,12 @@ public final class PCRoom {
 
     public var createdAtDate: Date { return self.dateFormatter.date(from: self.createdAt)! }
     public var updatedAtDate: Date { return self.dateFormatter.date(from: self.updatedAt)! }
+    public var deletedAtDate: Date? {
```
hamchapman: The type of this is different to the others. The force unwrapping I was doing with the others _shouldn't_ be problematic (given we know the format of timestamps that we're sending from the server), however, it does appear like there might be problems with it: https://github.com/pusher/chatkit-swift/issues/107

If we change one, we should change them all. Or not change any... I'd suggest following the format of the others and fixing any potential crash separately

callum-oakley: I am still force unwrapping the parse, the different signature is because `deletedAt` can be `nil`.

hamchapman: Ah, sorry. Rushed the review and it shows 🤦‍♂️ Merge away

PR #127, opened 2019-01-15
Reconnect reconcilliation

hamchapman:
### What?

Primarily:

* Make reconnection lead to delegate hooks being called if state has changed between disconnection and reconnection

Also:

* Makes SDK much more robust
* Makes tests much more robust

### Why?

* Make state consistent for all clients
* Make everything more stable

----

CC @hamchapman

### DIFF
--- a/Sources/ChatManager.swift
+++ b/Sources/ChatManager.swift
@@ -270,6 +270,7 @@ import NotificationCenter
 
     // TODO: Maybe we need some sort of ChatManagerConnectionState?
     public func disconnect() {
+        self.logger.log("Disconnect called", logLevel: .verbose)
         currentUser?.userSubscription?.end()
         currentUser?.userSubscription = nil
         currentUser?.presenceSubscription?.end()
--- a/Sources/ChatManager.swift
+++ b/Sources/ChatManager.swift
@@ -189,7 +189,7 @@ import NotificationCenter
                 guard roomsPayload.count > 0 else {
                     strongSelf.informConnectionCoordinatorOfCurrentUserCompletion(currentUser: strongSelf.currentUser, error: nil)
                     if alreadyHadCurrentUser {
-                        strongSelf.reconcileRoomsIfNecessary(
+                        strongSelf.reconcileRooms(
                             old: existingRooms,
                             new: [],
                             roomStore: strongSelf.currentUser!.roomStore,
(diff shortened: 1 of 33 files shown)


Sources/ChatManager.swift
```diff
@@ -259,6 +332,62 @@ import NotificationCenter
...
+    delegate: PCChatManagerDelegate
+) {
+    let oldSet = Set(old)
+    let newSet = Set(new)
+
+    let roomsRemovedFrom = oldSet.subtracting(newSet)
+    let roomsAddedTo = newSet.subtracting(oldSet)
+
+    roomsRemovedFrom.forEach { room in
+        let removedRoom = roomStore.removeSync(id: room.id)
```
callum-oakley: It feels a bit odd that `reconcileRooms` mutates the room store in the removed from room case, but the other cases are all handled elsewhere, and none of the other `reconcileX` functions mutate the stores.

Sources/ChatManager.swift
```diff
@@ -170,16 +180,30 @@ import NotificationCenter
...
+                var existingRooms: [PCRoom] = []
+
                 // If the currentUser property is already set then the assumption is that there was
                 // already a user subscription and so instead of setting the property to a new
                 // PCCurrentUser, we update the existing one to have the most up-to-date state
                 if let currentUser = strongSelf.currentUser {
+                    // We need to take copies of the rooms so that when we make the comparisons
+                    // of received rooms to pre-existing rooms we aren't actually just comparing
+                    // the same rooms (as PCRoom is a class and we have reference semantics)
+                    existingRooms = currentUser.rooms.map { $0.copy() }
```
callum-oakley: Could probably remove the need to take a copy by tracking the differences while you're updating the store. Not sure how much of a reshuffle this would necessitate. Maybe not worth it.

Sources/PCUserStoreCore.swift
```diff
@@ -10,22 +10,26 @@ public final class PCUserStoreCore {
     }
 
     func addOrMerge(_ user: PCUser) -> PCUser {
-        let insertResult = self.users.insert(user)
+        return self.userStoreQueue.sync {
```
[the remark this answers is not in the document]
hamchapman: Yep, exactly that

Sources/ChatManager.swift
hamchapman: Updated and was actually able to massively simplify the user subscription initial state handling logic that was there previously: https://github.com/pusher/chatkit-swift/pull/127/commits/f04ec06b759983cf9c492feb6ec35bef4cefd64f

So thanks!

Sources/ChatManager.swift
hamchapman: Don't think it's worth it for now

PR #148, opened 2019-02-14
V3 simple message send

vivangkumar:
### What?

Support simple message sends as part of the v3 API. Currently only inline and url payloads are supported.

### Why?

We'd like to support multipart messages and this is the first step to get there. Next steps are to be able to send attachments via multipart messages (a case which is currently skipped).


vivangkumar:
[[0.5] Send v3 message (inline messages only)](https://trello.com/c/p4ZYxjDO/1633-05-send-v3-message-inline-messages-only)

Sources/ChatManager.swift
```diff
@@ -108,7 +116,7 @@ import NotificationCenter
         self.basicCurrentUser = PCBasicCurrentUser(
             id: userID,
             pathFriendlyID: pathFriendlyUserID,
-            instance: instance,
+            instance: v2Instance,
```
callum-oakley: v3?

Sources/PCMultipartMessage.swift
```diff
@@ -0,0 +1,164 @@
...
+        self.name = name
+        self.customData = customData
+        self.refresher = refresher
+        self.refreshURL = refreshURL
+        self.downloadURL = downloadURL
+        self.expiration = expiration
+    }
+    
+    func url() -> String {
+        if (Date() > self.urlExpiry()) {
```
callum-oakley: I realise this is preparatory work (and I question it's place in this PR, but mike did this too so you're in good company 😉) but the spec suggests we should be refreshing the URL if it has less than 30 minutes of validity remaining.

> The url() method on AttachmentPayload must return a URL with at least 30 minutes of validity remaining.

Tests/MessageTests.swift
```diff
@@ -223,7 +223,53 @@ class MessagesTests: XCTestCase {
...
+                        XCTAssertEqual(messages!.last?.text, "hola!")
+                        ex.fulfill()
+                    })
+                }
+            )
+        }
+        
+        waitForExpectations(timeout: 15)
+    }
+
```
callum-oakley: Can we have a test with a URL part as well? (it should end up as an attachment when you fetch it, so we have everything we need to make sure it's working.)

Sources/PCCurrentUser.swift
```diff
@@ -554,7 +554,7 @@ public final class PCCurrentUser {
 
     // MARK: Message-related interactions
 
-    func sendMessage(_ messageObject: [String: Any], roomID: String, completionHandler: @escaping (Int?, Error?) -> Void) {
+    func sendMessage(instance: Instance, _ messageObject: [String: Any], roomID: String, completionHandler: @escaping (Int?, Error?) -> Void) {
```
callum-oakley: It weirds me out that this method takes `[String: Any]` when we have perfectly good message types, but that's probably a refactor for another day... 😛

vivangkumar: Ah yes! I know..I'm keeping it consistent with what is already there :P The library could do with a refactor for sure.

Tests/MessageTests.swift
vivangkumar: Sure! Good spot.

Sources/PCMultipartMessage.swift
vivangkumar: aaah! I was going along with Mike's android PR since this is the first bit of swift work I've done. I think I'm going to remove this so we can keep the PR limited to its scope. But will keep the promise comment in mind the next time I implement it.

Sources/ChatManager.swift
vivangkumar: Yep! good spot.

vivangkumar:
@callum-oakley addressed your comments and (force)pushed.

vivangkumar:
Ha, I consciously tried to keep them small and grouped.

PR #151, opened 2019-02-19
Subscribe v3

vivangkumar:
### What?

Allows subscription to v3 messages.

### Why?

To support multipart messages.

----


vivangkumar:
[[0.5] Subscribe to v3 message](https://trello.com/c/EiMhZYeL/1640-05-subscribe-to-v3-message)

Sources/PCMultipartMessageEnricher.swift
```diff
@@ -0,0 +1,153 @@
...
+                self.enrichedMessagesAwaitingCompletionCalls.removeValue(forKey: messageID)
+            } while self.completionOrderList.first != nil && self.enrichedMessagesAwaitingCompletionCalls[self.completionOrderList.first!] != nil
+        }
+    }
+}
+
+public enum PCMultipartMessageEnrichmentResult {
+    case success(PCMultipartMessage)
+    case error(Error)
+}
```
callum-oakley: There's a lot of duplication from `PCBasicMessageEnricher` here... and the enrichment process isn't conceptually any different since we don't care about the `parts` or the `text` in either case. Can we factor out the common stuff? Or even better make the enricher generic over both types? (Protocols?)

Sources/PCMultipartMessageSubscription.swift
```diff
@@ -0,0 +1,125 @@
...
+                
+                strongSelf.onIsTypingHook(room, user)
+            }
+        }
+    }
+    
+    func end() {
+        self.resumableSubscription.end()
+    }
+}
```
callum-oakley: Likewise, this is conceptually no different from `PCMessageSubscription`, it would be good to cut down on the duplication.

Sources/PCPayloadDeserializer.swift
```diff
@@ -192,6 +216,45 @@ struct PCPayloadDeserializer {
...
+            resultPartPayload = PCMultipartPayload.urlPayload(payload: PCMultipartURLPayload(type: partType, url: partURL!))
+        }
+        
+        if partAttachment != nil {
+            let multipartAttachment = try createMultipartAttachmentFromPayload(partAttachment!)
+            resultPartPayload = PCMultipartPayload.attachmentPayload(payload: multipartAttachment)
+        }
+        
+        return PCPart(payload: resultPartPayload)
+    }
```
callum-oakley: The variable assignments followed by the nil checks seems redundant. Can we just do the thing in the if block. e.g.

```swift
        if let url = payload["url"] as? String {
            return PCPart(payload: PCMultipartPayload.urlPayload(
                payload: PCMultipartURLPayload(type: partType, url: url)
            ))
        }
```

Sources/PCRoomDelegate.swift
```diff
@@ -3,6 +3,7 @@ import PusherPlatform
 
 public protocol PCRoomDelegate: AnyObject {
     func onMessage(_ message: PCMessage)
+    func onMultipartMessage(_ message: PCMultipartMessage)
```
callum-oakley: I'm not sure how I feel about `onMultipartMessage`, but I can see how the whole room delegate thing might force our hand on this one. Is @mdpye doing this in kotlin? It might be too weird if Swift is the only SDK with the new hook name. Can we avoid it? (I guess you might need two room delegate types, which is a bit icky... 🤔

callum-oakley:
also that CI bot is a disgrace and we should remove it until we're going to start enforcing its rules 😅

Sources/PCRoomDelegate.swift
vivangkumar: I tried implementing this like I discussed with @mdpye but because of the structure of the SDK it is definitely icky and it proved to be a much more deep rooted change than we initially thought :(

Sources/PCPayloadDeserializer.swift
vivangkumar: Yeah, the compiler will complain since we still have to return something at the end of the function that don't fall in any of the branches despite the fact that we're guaranteed that one of the branches will run. Not sure there's a nicer way to organise this bit..

Sources/PCMultipartMessageEnricher.swift
vivangkumar: Ahh I had this in the back of my mind as something to do at the end. Good you brought this up! Also at the same time I wasn't sure if I should start adding new things to the library or maintain the existing way of doing things..

vivangkumar: Although yeah, I think if I make this generic enough, then we can remove the other duplicated subscription you mentioned below..

Sources/PCPayloadDeserializer.swift
callum-oakley: Well maybe at least

```swift
        if let url = payload["url"] as? String {
            resultPartPayload = PCMultipartPayload.urlPayload(
                payload: PCMultipartURLPayload(type: partType, url: url)
            )
        }
```

then still return at the end

vivangkumar:
Ugh stupid Xcode. Removing it from Xcode just removes the file from the directory view but doesn't remove the actual file itself..

PR #155, opened 2019-02-27
Upload attachments v3

vivangkumar:
### What?

Allows attachment uploads using the new v3 api.

### Why?

To finish off multipart messages.

----

### DIFF
--- a/Sources/PCMultipartAttachmentUploader.swift
+++ b/Sources/PCMultipartAttachmentUploader.swift
@@ -19,7 +19,8 @@ class PCMultipartAttachmentUploader {
     fileprivate let uploadTasks: Array<PCMultipartAttachmentUploadTask>
     fileprivate let dispatchQueue = DispatchQueue(label: "com.pusher.chatkit.multipart-upload-\(UUID().uuidString)")
     fileprivate let dispatchGroup = DispatchGroup()
-    fileprivate let uploadResults = PCSynchronizedArray<PCMultipartAttachmentUploadResult>()
+    fileprivate let uploadSuccesses = PCSynchronizedArray<PCMultipartAttachmentUploadResult>()
+    fileprivate let uploadFailures = PCSynchronizedArray<Error>()
 
     init(
         instance: Instance,
@@ -29,21 +30,22 @@ class PCMultipartAttachmentUploader {
         self.uploadTasks = uploadTasks
     }
 
-    func getResults() throws -> [PCMultipartAttachmentUploadResult] {
+    func upload(completionHandler: @escaping ([PCMultipartAttachmentUploadResult]?, [Error]?) -> Void) {
         for task in self.uploadTasks {
             self.dispatchGroup.enter()
             self.dispatchQueue.async {
                 self.uploadMultipartAttachment(task: task) { (attachmentID, error) in
                     defer { self.dispatchGroup.leave() }
                     guard error == nil else {
+                        self.uploadFailures.append(error!)
                         self.instance.logger.log(
                             "Failed to upload multipart attachment: \(error.debugDescription)",
-                            logLevel: .debug
+                            logLevel: .error
                         )
                         return
                     }
 
-                    self.uploadResults.append(
+                    self.uploadSuccesses.append(
                         PCMultipartAttachmentUploadResult(
                             attachmentID: attachmentID!,
                             partNumber: task.partNumber,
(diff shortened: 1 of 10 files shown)


vivangkumar:
[[1] Send v3 message attachment](https://trello.com/c/IL6BLMFt/1645-1-send-v3-message-attachment)

Sources/PCMultipartMessage.swift
```diff
@@ -46,46 +46,134 @@ extension PCMultipartMessage: Hashable {
...
+        self.customData = customData
+        self.size = size
+        self.urlRefresher = urlRefresher
+        self.refreshUrl = refreshUrl
+        self.downloadUrl = downloadUrl
+        self.expiration = expiration
+    }
+
+    func url(completionHandler: @escaping (String?, Error?) -> Void) {
+        if (Date() > self.urlExpiry()) {
```
callum-oakley: need a 30 minute buffer here

> The url() method on AttachmentPayload must return a URL with at least 30 minutes of validity remaining. If the URL stored in the object has less than 30 minutes of validity remaining, the method must use the refresh URL to update it, and return the new value. The value returned from urlExpiry() must also be updated.

Tests/Helpers/Helpers.swift
```diff
@@ -169,6 +169,28 @@ func deleteInstanceResources(completionHandler: @escaping (TestHelperError?) ->
     }.resume()
 }
 
+func synchronousHTTPRequest(
```
callum-oakley: If it's synchronous, why not return `(Data?, URLResponse?, Error?)` directly? Having a synchronous method with a completionHandler feels weird.

Sources/PCMultipartAttachmentUploader.swift
```diff
@@ -0,0 +1,174 @@
...
+        instance: Instance,
+        uploadTasks: [PCMultipartAttachmentUploadTask]
+    ) {
+        self.instance = instance
+        self.uploadTasks = uploadTasks
+    }
+
+    func upload(completionHandler: @escaping ([PCMultipartAttachmentUploadResult]?, [Error]?) -> Void) {
+        for task in self.uploadTasks {
+            self.dispatchGroup.enter()
```
callum-oakley: You could use PCProgressCounter here? Not sure it simplifies things, so up to you.

Tests/MessageTests.swift
```diff
@@ -408,7 +414,510 @@ class MessagesTests: XCTestCase {
...
+                        ex.fulfill()
+                    }
+                }
+            )
+        }
+
+        waitForExpectations(timeout: 15)
+    }
+
+    func testMultipartMessageWithMultipleeAttachmentsRetrievedOnV3() {
```
callum-oakley: `Multiplee`

Sources/PCMultipartAttachmentUploader.swift
vivangkumar: Hmmm, I didn't realise this existed :D But I think I'm going to leave it be for now.

Tests/Helpers/Helpers.swift
vivangkumar: d'oh!  🤦‍♂️

Sources/PCMultipartMessage.swift
```diff
@@ -46,46 +46,134 @@ extension PCMultipartMessage: Hashable {
 }
 
 public struct PCPart {
+    let type: String
```
callum-oakley: Oh, one more thing. This should be part of the payload according to the spec. (where it was before was correct...)

vivangkumar:
@callum-oakley Changes have been made.

Sources/PCMultipartMessage.swift
```diff
@@ -60,32 +64,128 @@ public struct PCMultipartURLPayload {
...
+        self.customData = customData
+        self.size = size
+        self.urlRefresher = urlRefresher
+        self.refreshUrl = refreshUrl
+        self.downloadUrl = downloadUrl
+        self.expiration = expiration
+    }
+
+    func url(completionHandler: @escaping (String?, Error?) -> Void) {
+        if (Date().addingTimeInterval(-(30 * 60)) > self.urlExpiry()) {
```
callum-oakley:
```suggestion
        if (Date().addingTimeInterval(30 * 60) > self.urlExpiry()) {
```

no?

vivangkumar: yes 🤦‍♂️ my bad.

PR #159, opened 2019-03-04
Update dangerfile

lukabratos:
### What?

See title.

### Why?

Rules need to be discussed in order to make lint's output more valuable.
Removing it at this point as it produces too much noise.

----


vivangkumar:
🎉

PR #163, opened 2019-03-19
Use public access modifier where needed

vivangkumar:
### What?

Use `public` access modifier for structs and classes that are externally facing.

### Why?

Not specifying one defaults to internal and leads to unintended results.

----


caldrian:
@vivangkumar @mdpye 
shouldn't the method `PCMultipartAttachmentPayload.url(_:)` be public as well?
I cannot use the sample code for downloading the attachment url

PR #166, opened 2019-03-21
Deprecate non-multipart messaging methods

mdpye:
### What?

Deprecate

### Why?

Because they are deprecated, and we want to encourage people to move to multipart messaging.
Eventually we will change these methods to be multipart, but we didn't want to break compatibility immedately.

----


mdpye:
@vivangkumar could you do the honour of releasing later please?

PR #172, opened 2019-04-09
Remove user cursor sub

vivangkumar:
### What?

This PR removes the user cursors subscription in favour of changes made on the server and moves to using the new v4 api.

### Why?

To pave the way for unread counts.

----

PR #173, opened 2019-04-09
Add unreadCount and lastMessageAt properties to PCRoom

vivangkumar:
### What?

Add unread message count and last message sent at properties to rooms.

### Why?

Allows users to easily fetch unread counts for a room and to also sort rooms by last message sent.

**NOTE**: **Depends on #172 being merged first.**

### DIFF
--- a/Sources/PCRoom.swift
+++ b/Sources/PCRoom.swift
@@ -10,6 +10,8 @@ public final class PCRoom {
     public internal(set) var updatedAt: String
     public internal(set) var deletedAt: String?
     public internal(set) var customData: [String: Any]?
+    public internal(set) var unreadCount: Int?
+    public internal(set) var lastMessageAt: String?
 
     public internal(set) var subscription: PCRoomSubscription?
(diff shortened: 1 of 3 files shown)


Sources/PCRoom.swift
```diff
@@ -84,6 +96,18 @@ public final class PCRoom {
                 (self.customData != nil && room.customData != nil &&
                     (self.customData! as NSDictionary).isEqual(to: room.customData!)
                 )
+            ) &&
+            (
+                (self.unreadCount == nil && room.unreadCount == nil) ||
```
elverkilde: this looks fascinating. Why is it not sufficient to check `self.unreadCount == room.unreadCount`?

vivangkumar: You're right! I thought there might be some quirkiness with checking for equality on Optionals, but according to the implementation they work like they should. ie. The operands are equal if they are both nil, or if they are both not nil and the unwrapped values are equal.

Will change.

PR #178, opened 2019-04-29
Message deleted event

vivangkumar:
### What?

Support `message_deleted` event.

### Why?

New deletion endpoints (v5)

**NOTE**: **DO NOT MERGE YET**

----


Sources/ChatManager.swift
```diff
@@ -31,7 +31,7 @@ import NotificationCenter
         didSet {
             connectionCoordinator.logger = logger
             v2Instance.logger = logger
-            v4Instance.logger = logger
+            v5Instance.logger = logger
```
mdpye: It might be worth just calling this one `instance`, if it is going to track the "current" version, while v2 remains specific. Otherwise we'll be renaming this member a lot!

vivangkumar: :+1:

PR #180, opened 2019-05-23
Public initializers for payloads in multipart message 

realityworks:
### What?
In order to create messages locally outside the library we need to have public initializers. Otherwise they cannot be initialized due to internal accessibilty of those initializers

### Why?
The ability to setup and send messages internally without having to wait until the message is pushed through the server means we can have a "message in transit". If these initializers are internal, the entire messaging system needs to replicate the message structure. With these public initializers we create custom internal messages.

----


vivangkumar:
Sorry for the late response! This looks good! Thank you for the PR!

PR #181, opened 2019-05-28
Update dependency Pusher Beams to the latest version (2.0.2)

lukabratos:
### What?

Update dependency Pusher Beams to the latest version (2.0.2)

### Why?

Version 2.0.2 of the Pusher Beams SDK contains many improvements and bug fixes.

----


lukabratos:
Correct, I've tested complete functionality by using a Chatkit instance and debug console. I've created two users, added them into a private room and triggered few push notifications.

lukabratos:
The only thing that I've discovered now is incompatibility between the macOS versions between Chatkit and Beams.

Minimum supported macOS version for Chatkit is [`10.11`](https://github.com/pusher/chatkit-swift/blob/master/PusherChatkit.podspec#L19) and [`10.12`](https://github.com/pusher/push-notifications-swift/blob/master/PushNotifications.podspec#L16) for Beams.

I'll need to change this in the Beams SDK to support `10.11` before we merge this PR.

PR #184, opened 2019-06-18
Simplify concurrency

mdpye:
### What?

Remove async methods from SynchronizedArray

### Why?

Over the course of multiple changes, the synchronized array has now arrived at
the only reliable implementation - a wrapper around a shared datastructure
with all access serialised via a *non-concurrent* queue.

There is no point submitting tasks to this queue asynchronously and passing in
a completion handler to be invoked when the operation is complete - all we
achieve doing this is a proliferation of callbacks and, more importantly, the
moving of the completion handler invocations in to the critical section. We
actually lose what ever concurrency there may be in the callers.

### DIFF
--- a/Tests/RoomMembershipTests.swift
+++ b/Tests/RoomMembershipTests.swift
@@ -52,6 +52,30 @@ class RoomMembershipTests: XCTestCase {
 
     // MARK: Chat manager delegate tests
 
+    func testMembersAreAvailableImmediatelyAfterCreation() {
+        let roomCreatedEx = expectation(description: "room was created")
+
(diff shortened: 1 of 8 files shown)


Tests/RoomMembershipTests.swift
```diff
@@ -52,6 +52,30 @@ class RoomMembershipTests: XCTestCase {
 
     // MARK: Chat manager delegate tests
 
+    func testMembersAreAvailableImmediatelyAfterCreation() {
```
vivangkumar: Nice!

PR #188, opened 2019-07-01
User supplied room ids

vivangkumar:
### What?

Allows supplying room id when creating a room.

### Why?

----


Sources/ChatManager.swift
```diff
@@ -50,7 +50,7 @@ import NotificationCenter
 
         self.logger = logger
 
-        let sdkInfo = PPSDKInfo(productName: "chatkit", sdkVersion: "1.6.0")
+        let sdkInfo = PPSDKInfo(productName: "chatkit", sdkVersion: "1.7.0")
```
callum-oakley: sucks that you have to manually update this every time. I wonder if you can pull it out of the podspec? (out of scope for this PR, but would be cool if it isn't a lot of work)

Sources/ChatManager.swift
```diff
@@ -166,7 +166,7 @@ import NotificationCenter
                         id: strongSelf.userID,
                         pathFriendlyID: strongSelf.pathFriendlyUserID,
                         instance: strongSelf.instance,
-                        v5Instance: strongSelf.v5Instance,
+                        v5Instance: strongSelf.v6Instance,
```
callum-oakley: `v5instances`

Sources/PCCurrentUser.swift
```diff
@@ -81,7 +81,7 @@ public final class PCCurrentUser {
         self.avatarURL = avatarURL
         self.customData = customData
         self.instance = instance
-        self.v5Instance = v5Instance
+        self.v6Instance = v5Instance
```
callum-oakley: more v5

vivangkumar: Stupid Xcode..I used the refactor thing where it should change everything..but I don't think it quite worked..

vivangkumar:
Ok I think that should be it.

PR #197, opened 2019-08-06
Fix race condition related to PPRepeater

grzesiekko:
### What?

- Fix race condition related to the usage of `state` property of `PPRepeater`.
- Update dependencies.
- Clean up deployment target to match settings from the main target.

### Why?

Daisie report describing race conditions in `PPRepeater` class.

----


CHANGELOG.md
```diff
@@ -4,29 +4,35 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/pusher/chatkit-swift/compare/1.8.0...HEAD)
+## [Unreleased](https://github.com/pusher/chatkit-swift/compare/1.8.1...HEAD)
+
+## [1.8.1](https://github.com/pusher/chatkit-swift/compare/1.8.1...1.8.1) - 2019-08-06
```
lukabratos:
```suggestion
## [1.8.1](https://github.com/pusher/chatkit-swift/compare/1.8.0...1.8.1) - 2019-08-06
```

Cartfile
```diff
@@ -1,5 +1,5 @@
-github "pusher/pusher-platform-swift" ~> 0.7.1
-github "pusher/push-notifications-swift" ~> 2.0.2
+github "pusher/pusher-platform-swift" ~> 0.7.2
+github "pusher/push-notifications-swift" ~> 2.1.2
```
lukabratos: It should be all fine but I would update `push-notifications-swift` dependency in the separate commit.
Because this PR is fixing race condition with `PPRepeater`, if something potentially goes wrong we don't know if it's a change in the platform SDK or Beams SDK.

Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
 github "hamchapman/Mockingjay" "6bbe3ad2139609e56c0bd5f2fa0e7911dc13a6f8"
-github "krzyzanowskim/CryptoSwift" "0.13.1"
-github "pusher/push-notifications-swift" "2.0.2"
-github "pusher/pusher-platform-swift" "0.6.4"
+github "krzyzanowskim/CryptoSwift" "1.0.0"
+github "pusher/push-notifications-swift" "2.1.2"
```
lukabratos: See the comment above.

PusherChatkit.podspec
```diff
@@ -11,12 +11,12 @@ Pod::Spec.new do |s|
...
+  s.dependency 'PusherPlatform', '~> 0.7.2'
+  s.ios.dependency 'PushNotifications', '~> 2.1.2'
+  s.macos.dependency 'PushNotifications', '~> 2.1.2'
 
   s.ios.deployment_target = '10.0'
-  s.macos.deployment_target = '10.11'
-  s.tvos.deployment_target = '9.0'
-  s.watchos.deployment_target = '2.0'
+  s.macos.deployment_target = '10.12'
+  s.tvos.deployment_target = '10.0'
```
lukabratos: If the code works with older deployment targets I would leave it as it is.
If we're using a new API that is not included in the "old" version of dependency I would update them to a newer version.

PusherChatkit.xcodeproj/project.pbxproj
```diff
@@ -617,15 +613,15 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
lukabratos: Same as comment above.

PusherChatkit.xcodeproj/project.pbxproj
```diff
@@ -788,11 +778,9 @@
 				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos macosx appletvsimulator watchsimulator watchos appletvos";
 				SWIFT_VERSION = 4.2;
 				TARGETED_DEVICE_FAMILY = "1,2,3,4";
-				TVOS_DEPLOYMENT_TARGET = 10.0;
```
lukabratos: Were those deletions of deployment targets intentional?

Cartfile
grzesiekko: It is a separate commit 5147c155fafb62f1b158c12881e3d4f9005d5e19. Did you mean a separate pull request? I can submit it as a separate pull request. I just thought that because fixing `PPRepeater` is mostly a dependency update, I will also update other things.

Cartfile.resolved
grzesiekko: These two are only used by the test target and have no impact on the framework.

PusherChatkit.podspec
grzesiekko: This was set to incorrect values. The framework target is set to 10.12 for macOS and 10.0 for iOS while podspec file is manifesting compatibility with older versions.

PusherChatkit.xcodeproj/project.pbxproj
grzesiekko: This deletion is intentional. I moved these settings from target level to project level so that it is now consistent across all targets.

PusherChatkit.xcodeproj/project.pbxproj
grzesiekko: This setting is for the test target. The framework target was specifying 10.0.

PR #199, opened 2019-08-23
Public data model

grzesiekko:
### What?

- Public data model.
- Snapshotting mechanism for converting internal data model into the public one.
- Store providing functionality to fetch entities from Core Data.
- Data providers for accessing different parts of the public data model.

### Why?

We need to provide our users with ability to consume the data provided by our APIs.

----


Unit Tests/Tests/Model/MessageTests.swift
```diff
@@ -0,0 +1,420 @@
...
+    
+    func testShouldCreateMessageWithCorrectValues() {
+        let message = Message(identifier: "testIdentifier",
+                              sender: self.firstTestUser,
+                              parts: [self.testTextPart],
+                              readByUsers: [self.firstTestUser, self.secondTestUser, self.thirdTestUser],
+                              lastReadByUsers: [self.secondTestUser],
+                              createdAt: Date.distantPast,
+                              updatedAt: self.now,
+                              deletedAt: Date.distantFuture,
```
mdpye: I don't know if it affects anything, but to be realistic, deleteAt should either be `null`, or in the past. It should never be in the future.

Chatkit/Model/Message.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import CoreData
+
+public struct Message {
+    
+    public let identifier: String
```
mdpye: Everywhere else in the service (and pretty much in the world, as far as I can tell), this is abbreviated at `id` :)

Unit Tests/Tests/Model/MessageTests.swift
grzesiekko: It does not affect anything. It allows test to be stable and provide a constant value different from current date.

Chatkit/Model/Message.swift
grzesiekko: `id` is a reserved keyword in...Objective-C therefore it is not widely used on the platform.

mdpye: Oh, nice...

Unit Tests/Tests/Model/MessageTests.swift
mdpye: OK. Personally, I'd got for `null` then (if the message is not deleted), as the backend defines deleted to be `deletedAt != null`, not `deleteAt < now`.

grzesiekko: `testShouldCreateMessageWithCorrectValues()` <- it is hard to test value assignments with `nil` value. None of the values used in unit tests make much sens (intentionally) as that is not what those test are about. Real values should be used in integration tests.

Chatkit/Model/User.swift
```diff
@@ -0,0 +1,68 @@
...
+    public let presenceState: PresenceState
+    public let metadata: Metadata?
+    public let createdAt: Date
+    public let updatedAt: Date
+    
+    let objectID: NSManagedObjectID
+    
+    // MARK: - Initializers
+    
+    init(identifier: String, name: String?, avatar: URL?, presenceState: PresenceState, metadata: Metadata?, createdAt: Date, updatedAt: Date, objectID: NSManagedObjectID) {
```
lukabratos: I think that with `struct`'s we can even leave the `init` out as the compiler will create one for us.

Chatkit/Persistence/Model/Entities/InstanceEntity+CoreDataProperties.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+import CoreData
+
+extension InstanceEntity {
+    
+    @nonobjc class func fetchRequest() -> NSFetchRequest<InstanceEntity> {
```
lukabratos: Out of curiosity - what is the setting for `Codegen`? Are we manually writing NSManagedObjects or is it done automatically by using `Class Definition` setting?

Chatkit/Model/Room.swift
```diff
@@ -0,0 +1,83 @@
...
+    public let metadata: Metadata?
+    public let createdAt: Date
+    public let updatedAt: Date
+    public let deletedAt: Date?
+    
+    let objectID: NSManagedObjectID
+    
+    // MARK: - Initializers
+    
+    init(identifier: String, name: String?, isPrivate: Bool, creator: User, members: Set<User>?, typingMembers: Set<User>?, unreadCount: Int64, lastMessage: Message?, metadata: Metadata?, createdAt: Date, updatedAt: Date, deletedAt: Date?, objectID: NSManagedObjectID) {
```
lukabratos: We could remove this `init` too.

Chatkit/Model/Room.swift
```diff
@@ -0,0 +1,83 @@
...
+    
+}
+
+// MARK: - Equatable
+
+extension Room: Equatable {
+    
+    public static func == (lhs: Room, rhs: Room) -> Bool {
+        // Metadata is intentionally excluded from this comparison.
+        return lhs.identifier == rhs.identifier
```
lukabratos: If metadata is excluded; can we guarantee then that two objects will truly be equal then if we don't compare all properties?

Chatkit/Model/User.swift
```diff
@@ -0,0 +1,68 @@
...
+        hasher.combine(self.identifier)
+    }
+    
+}
+
+// MARK: - Equatable
+
+extension User: Equatable {
+    
+    public static func == (lhs: User, rhs: User) -> Bool {
```
lukabratos: It's something for the future but we could use [Sourcery](https://www.raywenderlich.com/501-sourcery-tutorial-generating-swift-code-for-ios) to auto-generate these things.

Unit Tests/Tests/Store/StoreTests.swift
```diff
@@ -0,0 +1,172 @@
+import XCTest
+import PusherPlatform
+@testable import PusherChatkit
+
+class StoreTests: XCTestCase {
```
lukabratos: Are we running these tests on the CI at the moment?
There is just one check but it is related to Synk.

Chatkit/Model/User.swift
grzesiekko: That is true. However, there is one nasty side effect of using auto-generated initializers in Swift. Changing the order of properties in the file changes the order of parameters in the initializer.

Example:

#### Variant A

```swift
public struct User {
    public let identifier: String
    public let name: String
}
```

Generates:

```swift
public init(identifier: String, name: String) {
    self.identifier = identifier
    self.name = name
}
```

#### Variant B

```swift
public struct User {
    public let name: String
    public let identifier: String
}
```

Generates:

```swift
public init(name: String, identifier: String) {
    self.name = name
    self.identifier = identifier
}
```

This is a subtle change, but it has the potential to break the public interface. That is why I would prefer to be extra cautious and generate initializers manually.

There is also one more problem here. Auto-generated initializer would be defined as `public` and we do not want anyone to instantiate our model outside our SDK.

Chatkit/Persistence/Model/Entities/InstanceEntity+CoreDataProperties.swift
grzesiekko: It is set to `Manual/None`. Both classes and extensions are generated by Xcode and amended manually for nullability (which is unfortunate, because Xcode should be able to figure this out automatically from the model file).

We can `mogenerator` for that if that is a problem.

Chatkit/Model/Room.swift
grzesiekko: Please see my replay above (related to `User.swft` file).

Chatkit/Model/Room.swift
grzesiekko: No we cannot. The problem is that `Metadata` is defined as `[String : Any]` and `Any` does not implement `Equatable` in Swift.

We could try to use the following approach to solve this problem: https://stackoverflow.com/questions/47632263/comparing-two-string-any-dictionaries-in-swift-4?rq=1

Chatkit/Model/User.swift
grzesiekko: Sure, that an option for us to consider.

Unit Tests/Tests/Store/StoreTests.swift
grzesiekko: I think we are not and that is because there is workspace anymore. We can change that now, but that might affect running tests on `master`.

PR #202, opened 2019-09-09
List of rooms

grzesiekko:
### What?

Logic required to connect and fetch a list of available rooms for the current user.

### Why?

https://pusherpm.leankit.com/card/888673523

----


Unit Tests/Tests/Chatkit/ChatkitTests.swift
```diff
@@ -1,23 +1,130 @@
...
+    
+    func testShouldReturnNoErrorAfterSuccessfulConnection() {
+        stubSubscription(of: .chat, version: .version6, instanceLocator: Networking.testInstanceLocator, path: .users, with: "chat-initial_state")
+        
+        let chatkit = try? Chatkit(instanceLocator: Networking.testInstanceLocator, tokenProvider: TestTokenProvider())
+        
+        let expectation = self.expectation(description: "Connection")
+        
+        chatkit?.connect { error in
+            XCTAssertNil(error)
```
lukabratos: Would it be beneficial if we would check also the HTTP response code and error message?

Unit Tests/Tests/Networking/SDK Info/PPSDKInfoExtensionsTests.swift
```diff
@@ -0,0 +1,34 @@
...
+class PPSDKInfoExtensionsTests: XCTestCase {
+    
+    // MARK: - Tests
+    
+    func testShouldHaveCurrentSDKInfo() {
+        XCTAssertNotNil(PPSDKInfo.current)
+    }
+
+    // FIXME: Uncomment this test after changing access level of PPSDKInfo properties in PusherPlatform SDK.
+//    func testCurrentSDKInfoShouldHaveProductNameSetToChatkit() {
```
lukabratos: Should we remove commented code out of this commit? We can add this in later when it will be needed.

Chatkit/Networking/Client/PCBaseClient+Extensions.swift
```diff
@@ -0,0 +1,26 @@
...
+    // MARK: - Internal methods
+    
+    class func host(for instanceLocator: String) throws -> String {
+        let components = instanceLocator.split(separator: PPBaseClient.instanceLocatorSeparator)
+        
+        guard components.count == 3 else {
+            throw NetworkingError.invalidInstanceLocator
+        }
+        
+        let cluster = components[1]
```
lukabratos: I was thinking if we could use `enum` to describe components instead of numbers but it might be overkill.

Unit Tests/Tests/Chatkit/ChatkitTests.swift
grzesiekko: Platform SDK does not provide us with such information.

Unit Tests/Helpers/XCTest+Networking.swift
```diff
@@ -0,0 +1,73 @@
...
+internal extension XCTest {
+    
+    // MARK: - Properties
+    
+    private static let instanceLocatorSeparator: Character = ":"
+    
+    // MARK: - Internal methods
+    
+    @discardableResult func stubSubscription(of service: ServiceName, version: ServiceVersion, instanceLocator: String, path: Networking.Path, with jsonFilename: String) -> Mockingjay.Stub? {
+        let bundle = Bundle(for: type(of: self))
```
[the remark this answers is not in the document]
grzesiekko: Unfortunately, we cannot. Currently `Bundle.current` is hardcoded to `PusherChatkit` bundle and here we need to get `Unit Tests` bundle. I initially wanted to get rid of the hardcoded value and make this dynamic. It turns out that such thing (relatively easy to do in Objective-C) would require >1000 lines of Swift code and even than it would not return the expected value in 100% situations.

lukabratos:
It might be part of the next PR but we should add comments to all of the `public` classes and structs so we can generate documentation later + it will help out customers as they will be using SDK in their code: https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/index.html

Chatkit/Utilities/Bundle+Extensions.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+internal extension Bundle {
```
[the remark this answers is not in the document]
grzesiekko: Extension to public class is by default public.

> You can extend a class, structure, or enumeration in any access context in which the class, structure, or enumeration is available. Any type members added in an extension have the same default access level as type members declared in the original type being extended. If you extend a public or internal type, any new type members you add have a default access level of internal. If you extend a file-private type, any new type members you add have a default access level of file private. If you extend a private type, any new type members you add have a default access level of private.

https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html (This is for Swift 5.1 bu it is the same in Swift 4.2).

Unit Tests/Tests/Networking/SDK Info/PPSDKInfoExtensionsTests.swift
grzesiekko: This is a valid test that should be there. In my opinion getting rid of this now would effectively mean that we will never write that unit test in future (we will have more important tasks to do).

grzesiekko:
> It might be part of the next PR but we should add comments to all of the `public` classes and structs so we can generate documentation later + it will help out customers as they will be using SDK in their code: https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/index.html

Sure we plan to do that. It's one of the epics for SDK 2.0.

Chatkit/Networking/Client/PCBaseClient+Extensions.swift
grzesiekko: Well, we can do that or we can improve Platform SDK to handle this value in a more structured way.

PR #205, opened 2019-10-23
Chatkit semantics

grzesiekko:
### What?

Chatkit semantics.

### Why?

Apparently someone might want to use that.

----


mdpye:
I suspect the paged providers should have delegated `didReceive...` rather than `didAdd...`.

Otherwise I think it looks good.

grzesiekko:
I will amend that.

PR #206, opened 2019-10-24
Typing users view model

grzesiekko:
### What?

Typing users view model.

### Why?

.

----


Chatkit/View Models/TypingUsersViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+    
+    // MARK: - Private methods
+    
+    private func reload() {
+        guard self.provider.typingUsers.count > 0 else {
+            self.value = nil
+            return
+        }
+        
+        let names = self.provider.typingUsers.map { $0.name ?? "x" }.sorted().joined(separator: ", ")
```
mdpye: `"x"` -> `self.userNamePlaceholder`?

Chatkit/View Models/TypingUsersViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+/// A view model which provides a text representations of typing users.
+public class TypingUsersViewModel {
+    
+    // MARK: - Properties
+    
+    /// The text representations of typing users.
+    public private(set) var value: String?
+    
+    /// The typing users provider used as the source of data.
+    public let provider: TypingUsersProvider
```
mdpye: Surely this is private? What's the benefit in exposing it?

Chatkit/View Models/TypingUsersViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+    
+    /// The placeholder used when a user doeas not have a value set for the name property.
+    public let userNamePlaceholder: String
```
mdpye: Surely this is private? What's the benefit in exposing it?

Chatkit/View Models/TypingUsersViewModel.swift
grzesiekko: Well it has been injected into the class, it is being retained by the class, so it make sense for this to be exposed as `public`. It is consistent with other SDKs. It also improves testability.

Chatkit/View Models/TypingUsersViewModel.swift
grzesiekko: Please see the comment above.

Chatkit/View Models/TypingUsersViewModel.swift
grzesiekko: Fixed.

Chatkit/View Models/TypingUsersViewModel.swift
grzesiekko: After the discussion the property is now marked as `private`.

Chatkit/View Models/TypingUsersViewModel.swift
grzesiekko: After the discussion the property is now marked as `private`.

PR #207, opened 2019-10-28
Joined rooms view model

grzesiekko:
### What?

Joined rooms view model.

### Why?

User testing.

### Other comments

View model might not be the best name fot this class.

----


Chatkit/View Models/JoinedRoomsViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+        self.rooms[previousIndex] = room
+        self.sort()
+        
+        guard let currentIndex = self.index(of: room) else {
+            return
+        }
+        
+        if let currentLastMessage = room.lastMessage,
+            let previousLastMessage = previousValue.lastMessage,
+            currentLastMessage.createdAt == previousLastMessage.createdAt {
```
mdpye: I suspect this condition is backwards. If the current and previous values of "last message at" are the same, then the event is "message received"?

Chatkit/View Models/JoinedRoomsViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+            return
+        }
+        
+        self.delegate?.joinedRoomsViewModel(self, didAddRoomAtIndex: index, changeReason: .userJoined)
+    }
+    
+    public func joinedRoomsProvider(_ joinedRoomsProvider: JoinedRoomsProvider, didUpdateRoom room: Room, previousValue: Room) {
+        // TODO: Optimize if necessary
+        
+        guard let previousIndex = self.index(of: previousValue) else {
```
mdpye: It would probably be safer (and less confusing) to find rooms in these lists by comparing their IDs, rather than comparing entire objects and trying to ensure we search using the correct previous/current instance.

Chatkit/View Models/JoinedRoomsViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+        self.sort()
+        
+        guard let currentIndex = self.index(of: room) else {
+            return
+        }
+        
+        if let currentLastMessage = room.lastMessage,
+            let previousLastMessage = previousValue.lastMessage,
+            currentLastMessage.createdAt == previousLastMessage.createdAt {
+            self.delegate?.joinedRoomsViewModel(self, didUpdateRoomAtIndex: currentIndex, changeReason: .messageReceived(previousIndex))
```
mdpye: If these are supposed to map closely to rendering cells in the collection view, perhaps this should be grouped update and move callbacks, rather than one which represents both operations?

Chatkit/View Models/JoinedRoomsViewModel.swift
grzesiekko: That is exactly what `index(of:)` method is doing.

Test/Controllers/RoomListViewController.swift
```diff
@@ -31,95 +31,83 @@ class RoomListViewController: UITableViewController {
...
         let indexPath = IndexPath(row: index, section: 0)
-        self.tableView.reloadRows(at: [indexPath], with: .fade)
+        
+        switch changeReason {
+        case .dataUpdated:
+            self.tableView.reloadRows(at: [indexPath], with: .fade)
+            
+        case let .messageReceived(previousIndex):
+            let previousIndexPath = IndexPath(row: previousIndex, section: 0)
+            self.tableView.moveRow(at: previousIndexPath, to: indexPath)
```
mdpye: Does moving a row automatically reload the data for it too?

grzesiekko: Yes, it does reload the content of the cell.

Chatkit/View Models/JoinedRoomsViewModel.swift
grzesiekko: Sure, we can have 2 delegate callbacks if that is our preference here.

Chatkit/View Models/JoinedRoomsViewModel.swift
grzesiekko: Yes, the `changeReason` is incorrect here.

Chatkit/View Models/JoinedRoomsViewModel.swift
mdpye: So it is, sorry!

mdpye: I guess my confusion was the explicit use of the old and new values when we know they have the same identifier.

PR #212, opened 2019-11-19
Plan names

grzesiekko:
### What?

Plan names added to data simulation.

### Why?

User testing.

----

### DIFF
--- a/Chatkit/Data Simulation/DataSimulator+Events.swift
+++ b/Chatkit/Data Simulation/DataSimulator+Events.swift
@@ -46,37 +46,37 @@ extension DataSimulator {
                 let eleventhUser = self.createUser(in: context, identifier: "amelia", name: "Amelia")
                 self.eleventhUserID = eleventhUser.objectID
                 
-                let firstRoom = self.createRoom(in: context, identifier: "firstRoom", userName: "Oliver", members: [currentUser, firstUser])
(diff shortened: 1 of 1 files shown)


Chatkit/Data Simulation/DataSimulator+Events.swift
```diff
@@ -192,17 +192,33 @@ extension DataSimulator {
...
+            
+            room.name = "\(userName) - \(planName) Plan"
+            room.userData = UserDataSerializer.serialize(userData: userData)
+        }
+        else {
+            let planName = "Nutrition"
+            let userData = ["planID" : "routine-basic",
+                            "planName" : planName]
+            
+            room.name = "\(userName) - \(planName) Plan"
```
mdpye: I think it would be more realistic to have just the `planName` in the room name field, and have the implementor combine the user name and plan name to come up with the display string. We expect dynamically computed conversation titles to be a common case, especially in one-to-one use case.

grzesiekko: No problem. Should we drop user name from the name of the room as well?

grzesiekko: No problem.

grzesiekko: Done

PR #213, opened 2019-11-19
Limit typing indicators

grzesiekko:
### What?

- Typing indicators displayed by `TypingUsersViewModel` limited to 3.
- Data simulation no longer sends typing indicators of the currently logged in user.

### Why?

User testing.

----

### DIFF
--- a/Chatkit/Data Providers/TypingUsersProvider.swift
+++ b/Chatkit/Data Providers/TypingUsersProvider.swift
@@ -27,13 +27,13 @@ public class TypingUsersProvider {
         return Set(users)
     }
     
-    let currentUser: User
+    let currentUserIdentifier: String
     
     // MARK: - Initializers
     
     init(room: Room, currentUser: User, persistenceController: PersistenceController) {
         self.roomIdentifier = room.identifier
-        self.currentUser = currentUser
+        self.currentUserIdentifier = currentUser.identifier
         self.state = .connected
         
         self.roomManagedObjectID = room.objectID
--- a/Chatkit/Data Providers/TypingUsersProvider.swift
+++ b/Chatkit/Data Providers/TypingUsersProvider.swift
@@ -27,13 +27,10 @@ public class TypingUsersProvider {
         return Set(users)
     }
     
-    let currentUserIdentifier: String
-    
     // MARK: - Initializers
     
-    init(room: Room, currentUser: User, persistenceController: PersistenceController) {
+    init(room: Room, persistenceController: PersistenceController) {
         self.roomIdentifier = room.identifier
-        self.currentUserIdentifier = currentUser.identifier
         self.state = .connected
         
         self.roomManagedObjectID = room.objectID
--- a/Chatkit/Data Providers/TypingUsersProvider.swift
+++ b/Chatkit/Data Providers/TypingUsersProvider.swift
@@ -27,10 +27,13 @@ public class TypingUsersProvider {
         return Set(users)
     }
     
+    let currentUser: User
+    
     // MARK: - Initializers
     
-    init(room: Room, persistenceController: PersistenceController) {
+    init(room: Room, currentUser: User, persistenceController: PersistenceController) {
         self.roomIdentifier = room.identifier
+        self.currentUser = currentUser
         self.state = .connected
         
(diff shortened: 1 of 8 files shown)


grzesiekko:
Actually, I might add 1 thing here before merging.

grzesiekko:
Done. The PR can be reviewed now.

mdpye:
Hah, I think you've really hit on something here!

I believe that you do receive your own typing indicators from the web service, so rather than removing them from the dummy data, we should probably be filtering the current user from the list in the view model...

grzesiekko:
I will add such functionality.

Chatkit/Data Providers/TypingUsersProvider.swift
```diff
@@ -27,10 +27,13 @@ public class TypingUsersProvider {
...
     }
     
+    let currentUser: User
+    
     // MARK: - Initializers
     
-    init(room: Room, persistenceController: PersistenceController) {
+    init(room: Room, currentUser: User, persistenceController: PersistenceController) {
         self.roomIdentifier = room.identifier
+        self.currentUser = currentUser
```
mdpye: As the other fields of user (name, custom data etc) might change, and won't be updated here, perhaps we should only store the id, as it's all we actually need to identify the user?

mdpye: Also, surely this should be registered in the ViewModel where it is used, not here?

If you're worried about the constructor args for the view model, I think it's either valid to require it, or we can add factory methods for view models to the Chatkit object as Zan suggested

grzesiekko: One could argue that the same is valid for `typingUsers` property of `TypingUsersProvider`.

mdpye: Aren't the typing users snapshot from the coredata frequently, at least? (but yes, that is a conversation we still need to have!)

Chatkit/Data Providers/TypingUsersProvider.swift
```diff
@@ -27,10 +27,13 @@ public class TypingUsersProvider {
         return Set(users)
     }
     
+    let currentUserIdentifier: String
```
mdpye: I still don't see that this at belonging in here when it's not used here

grzesiekko: OK, in that case should we instantiate the view model with an instance of `Chatkit` in order to be sure that the user passed to the view model is actually the one that is currently logged into the system?

grzesiekko: It look like we have 3 options here:

1. Leave it as it is right now and hide the `currentUser` from developers:
   - The public interface is simpler.
   - It is not possible to instantiate the view model with an incorrect instance of `User`.
   - Unfortunately, the identifier needs to be passed here somehow (internally).
2. Instantiate the view model with an instance of `Chatkit`:
   - The public interface is a bit more complex simpler.
   - It is still not possible to instantiate the view model with an incorrect instance of `User` (that's a good thing).
   - It **is** possible to instantiate the view model with an incorrect instance of `Chatkit`.
   - Initializer of the view model will have to be be optional (error handling would be optional as well).
3. Instantiate the view model with an instance of `User` or an identifier of the current user (`String`):
   - The public interface is a bit more complex simpler.
   - It **is** possible to instantiate the view model with an incorrect instance of `User` (that's not a good thing).
   - A bit cleaner implementation (internally).

mdpye: Hrm. I see what you mean. Personally, I'd definitely request it from the TypingIndicatorsViewModel constructor - I was even thinking to turn it in to a generic "filter user" parameter. I'm unsure about the correctness / ease of use tradeoff here.

There is a 4th option, which Zan suggested:
- create view models *from* the Chatkit object, like we do with providers.

I think this is definitely the most ergonomic, but it also makes it seem more "magic" to the customer, which would discourage them (or make a different way of implementing at least) their own view models in future.

grzesiekko: After the recent change all view models are now being initialized using factory methods.

PR #215, opened 2019-11-21
Stick to bottom (in demo)

mdpye:

### DIFF
--- a/Demo/Controllers/MessageViewController.swift
+++ b/Demo/Controllers/MessageViewController.swift
@@ -9,7 +9,12 @@ class MessageViewController: UIViewController {
     var chatkit: Chatkit?
     var messagesViewModel: MessagesViewModel?
     var typingUsersViewModel: TypingUsersViewModel?
-    
+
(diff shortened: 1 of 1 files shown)


Demo/Controllers/MessageViewController.swift
```diff
@@ -92,22 +97,24 @@ class MessageViewController: UIViewController {
...
+
+        let rowCount = self.tableView.numberOfRows(inSection: 0)
+        self.lastRowVisibleBeforeUpdate = lastVisibleRow == rowCount-1
+    }
+
+    private func afterContentChange() {
+        let rowCount = self.tableView.numberOfRows(inSection: 0)
+        let indexPath = IndexPath(row: rowCount-1, section: 0)
+        if (self.lastRowVisibleBeforeUpdate) {
+            self.tableView.scrollToRow(at: indexPath, at: .bottom, animated: false)
```
grzesiekko: In order to enable animations here and make the whole transition smooth we would need improve layout constraints.

PR #216, opened 2019-11-22
Prevent attempts to both remove and update a row in a single batch

mdpye:
When loading older messages, we were running an delete and an update on
a row with the same index within the same batch, when we added a message
at i, removed the date header at i+1 and then tried to update the
grouping of i+2, which now has index i+1 because of the removal of the
date header which was there. This crashes the app.

We now split the update batches up in to different operations for
safety, and do the bulk insert, the removal of the date header and the
fixing of the grouping in separate operations. The animations all still
overlap in time for a smooth visual transition.

The bracketing willUpdate and didUpdate callbacks are also moved in to a
function which takes a closure to execute between them so that the start
and end can be observed at the lexical block level.

### DIFF
--- a/Chatkit/View Models/MessagesViewModel.swift
+++ b/Chatkit/View Models/MessagesViewModel.swift
@@ -58,12 +58,16 @@ public class MessagesViewModel {
             
             return
         }
-        
-        self.addLoadingIndicator()
-        
+
+        self.batchViewUpdate {
+            self.addLoadingIndicator()
+        }
+
         self.provider.fetchOlderMessages(numberOfMessages: numberOfMessages) { error in
             if error != nil {
-                self.removeLoadingIndicator()
+                self.batchViewUpdate {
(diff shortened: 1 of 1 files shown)


mdpye:
@grzesiekko you did say that manually updating tables was dangerous :)

grzesiekko:
I am aware of this issue and a some other edge cases that I decided not to cover as this whole view model is a throwaway code anyway. In order to make date headers working correctly with batches of messages that stretch across multiple days we should split the incoming data into parts that should be inserted before and/or after the date header.

If it is really important to implement test data structured in such way, I suggest merging this PR without a code review as I am not able to asses the impact of the provided "block-based batch update" approach on all possible update scenarios.

mdpye:
> In order to make date headers working correctly with batches of messages that stretch across multiple days we should split the incoming data into parts that should be inserted before and/or after the date header.

Sounds a sensible solution.

>

PR #220, opened 2019-11-25
Split groups when one message is sent > 60s after the previous

mdpye:


grzesiekko:
Shouldn't we also update date headers in order to make this feature working?

mdpye:
It's only the bubble grouping. In order to show it off, we should add timestamps to either the first or last messages in a group.

I attempted that, but it's beyond by UIKit skills to make that label appear in a good place!

PR #223, opened 2019-11-27
Timestamps added to Demo app

grzesiekko:
### What?

Timestamps added to Demo app

### Why?

User testing

----

### DIFF
--- a/Demo/Extensions/DateFormatter+Extensions.swift
+++ b/Demo/Extensions/DateFormatter+Extensions.swift
@@ -13,7 +13,7 @@ internal extension DateFormatter {
     
     static let time: DateFormatter = {
         let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "HH:mm"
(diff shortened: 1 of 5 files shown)


Demo/Extensions/DateFormatter+Extensions.swift
```diff
@@ -0,0 +1,21 @@
...
+    static let header: DateFormatter = {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "d MMMM yyyy"
+        
+        return dateFormatter
+    }()
+    
+    static let time: DateFormatter = {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "HH:mm"
```
mdpye: Shall we include seconds, even if just because lots of our test data arrives in a short space of time?

grzesiekko: Done

PR #224, opened 2019-11-28
Users for room

mdpye:
### What?

Allow synchronous fetch of static list of room members

### Why?

So that we can display member names in room titles.


Chatkit/Chatkit/Chatkit.swift
```diff
@@ -289,6 +289,26 @@ public class Chatkit {
...
+        let predicate = NSPredicate(format: "ANY %K == %@", #keyPath(UserEntity.room), roomManagedObjectID)
+        let sortDescriptor = NSSortDescriptor(key: #keyPath(UserEntity.identifier), ascending: true) { (lhs, rhs) -> ComparisonResult in
+            guard let lhsString = lhs as? String, let lhs = Int(lhsString), let rhsString = rhs as? String, let rhs = Int(rhsString) else {
+                return .orderedSame
+            }
+            
+            return NSNumber(value: lhs).compare(NSNumber(value: rhs))
+        }
+
+        let memberEntities = context.fetchAll(UserEntity.self, withRelationships: nil, limit: 999, sortedBy: [sortDescriptor], filteredBy: predicate)
```
grzesiekko:
```suggestion
        let memberEntities = context.fetchAll(UserEntity.self, withRelationships: nil, sortedBy: [sortDescriptor], filteredBy: predicate)
```

grzesiekko: `limit` set to `0` means that there is no limit.

Chatkit/Chatkit/Chatkit.swift
```diff
@@ -289,6 +289,26 @@ public class Chatkit {
...
+    /// Retrieves a static snapshot of the `User`s who are currently members of the `Room`.
+    ///
+    /// - Parameters:
+    ///     - `room`: The `Room` for which members should be retrieved.
+    public func members(for room: Room) -> [User] {
+        let roomManagedObjectID = room.objectID
+        
+        let context = persistenceController.mainContext
+        let predicate = NSPredicate(format: "ANY %K == %@", #keyPath(UserEntity.room), roomManagedObjectID)
+        let sortDescriptor = NSSortDescriptor(key: #keyPath(UserEntity.identifier), ascending: true) { (lhs, rhs) -> ComparisonResult in
```
grzesiekko: At some point I will improve this numeric search. It is used in many places in the code, but that does not concern this PR.

Chatkit/Chatkit/Chatkit.swift
grzesiekko: There is one problem here. This code assumes that the method will be always called from the main thread, because `mainContext` is confined to that thread. That might not always be the case. In order to fix that we can wrap this code into `performAndWait` closure.

Demo/Controllers/MessageViewController.swift
```diff
@@ -17,14 +17,22 @@ class MessageViewController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        
+
+        guard let chatkit = self.chatkit else {
+            print("Error: MessageViewController view did load, but Chatkit is not available")
```
grzesiekko: Printing to console has a significant performance impact for the app (it affects a few journals/logs maintained by the OS). We should avoid that.

Demo/Controllers/MessageViewController.swift
```diff
@@ -17,14 +17,22 @@ class MessageViewController: UIViewController {
...
+            return
+        }
         guard let room = self.room else {
+            print("Error: MessageViewController view did load, but no room has been specified")
```
grzesiekko: Same here.

Demo/Controllers/MessageViewController.swift
grzesiekko: `fatalError` might be useful here.

Chatkit/Chatkit/Chatkit.swift
mdpye: Cool. I did wonder when I wrote it. It's a shame that `performAndWait` returns `()` rather than a generic. 🤷‍♂

PR #228, opened 2019-12-06
Auth interface

mdpye:
More docs required, but interested in initial feedback @grzesiekko @marekoid

### DIFF
--- a/Chatkit/Auth/TokenProvider.swift
+++ b/Chatkit/Auth/TokenProvider.swift
@@ -11,14 +11,14 @@ extension String: Error { }
 public typealias TokenProvider = PPTokenProvider
 public typealias TokenProviderResult = PPTokenProviderResult
 
-/// ChatkitTestTokenProvider retrieves tokens from the Chatkit service's Test Token Provider, which
+/// TestTokenProvider retrieves tokens from the Chatkit service's Test Token Provider, which
 /// is for development user only, and must be enabled for your instance in the Chatkit Dashboard.
 ///
 /// The test token provider will always sign a token for the requested userID, without applying any
 /// form of authentication.
-public class ChatkitTestTokenProvider: TokenProvider {
+public class TestTokenProvider: TokenProvider {
     
-    private let delegate: HTTPSTokenProvider
+    private let delegate: DefaultTokenProvider
     
     /// - Parameters:
     ///     - instanceLocator: The locator for your instance, the same value from the Dashboard
@@ -32,7 +32,7 @@ public class ChatkitTestTokenProvider: TokenProvider {
         let instanceId = "UNIMPLEMENTED"
         let path = "/services/chatkit_token_provider/v1/\(instanceId)/token"
         
-        self.delegate = HTTPSTokenProvider(method: "POST",
+        self.delegate = DefaultTokenProvider(method: "POST",
                                            host: host,
                                            path: path,
                                            headers: nil,
@@ -56,14 +56,14 @@ public enum KeyValueResult {
     
 }
 
-/// HTTPCallTokenProvider makes calls to a specified HTTP endpoint and expects to receive a token
+/// HTTPCallTokenProvider makes calls to a specified HTTPS endpoint and expects to receive a token
 /// from it.
 ///
 /// This is the implementation we recommend for production use, to request tokens from your backend
 /// system.
 ///
 /// If this class does not fit your needs, you can implement the `TokenProvider` protocol yourself.
-public class HTTPSTokenProvider: TokenProvider {
+public class DefaultTokenProvider: TokenProvider {
     
     private let method: String
     private let host: String
@@ -71,7 +71,7 @@ public class HTTPSTokenProvider: TokenProvider {
     private let getHeaders: AsyncKeyValueCall?
     private let getQueryParams: AsyncKeyValueCall?
     
-    /// Create an HTTPSTokenProvider which presents either headers or query parameters as part of
+    /// Create an DefaultTokenProvider which presents either headers or query parameters as part of
     /// the request. These should be used to identify your application user session to your backend
     /// so that it can issue a token for the user.
     ///
@@ -93,11 +93,11 @@ public class HTTPSTokenProvider: TokenProvider {
         self.init(method: method,
                   host: host,
                   path: path,
-                  getHeaders: HTTPSTokenProvider.wrapCredentialsValue(headers),
-                  getQueryParams: HTTPSTokenProvider.wrapCredentialsValue(queryParams))
+                  getHeaders: DefaultTokenProvider.wrapCredentialsValue(headers),
+                  getQueryParams: DefaultTokenProvider.wrapCredentialsValue(queryParams))
     }
     
-    /// Create an HTTPSTokenProvider which presents either headers or query parameters as part of
+    /// Create an DefaultTokenProvider which presents either headers or query parameters as part of
     /// the request. These should be used to identify your application user session to your backend
     /// so that it can issue a token for the user.
     ///
--- a/Chatkit/Auth/TokenProvider.swift
+++ b/Chatkit/Auth/TokenProvider.swift
@@ -20,22 +20,28 @@ public typealias TokenProviderResult = PPTokenProviderResult
 
 /// A factory for `TokenProvider`s.
 ///
-/// The methods on this object create token provider instances which can interface with an HTTP endpoint in your backend, or the Chatkit Test Token Provider (for use during development)
+/// The methods on this object create token provider instances which can interface with an HTTP 
+/// endpoint in your backend, or the Chatkit Test Token Provider (for use during development)
 public struct ChatkitTokenProviders {
-    /// Creates a `TokenProvider` which fetches tokens from the Chatkit Test Token Provider. It should be used for development purposes only.
+    /// Creates a `TokenProvider` which fetches tokens from the Chatkit Test Token Provider. It
+    /// should be used for development purposes only.
     ///
-    /// The Chatkit Test Token Provider is a token provider endpoint hosted by Pusher which returns tokens for the requested `userId` without authenticating the user. It is intended for development convenience and *must not be enabled in instances which host production data*.
+    /// The Chatkit Test Token Provider is a token provider endpoint hosted by Pusher which returns
+    /// tokens for the requested `userId` without authenticating the user. It is intended for
+    /// development convenience and *must not be enabled in instances which host production data*.
     ///
     /// The test token provider must be enabled for your instance in the Chatkit dashboard.
     ///
     /// - Parameters:
-    ///     - instanceLocator: The locator string for your instance, which you also provide when initialising the `Chatkit` object.
-    ///     - userId: The user identifier to fetch a token for. The provider will sign a valid token for any user identifier requested, without applying any kind of authentication.
+    ///     - instanceLocator: The locator string for your instance, which you also provide when
+    ///     initialising the `Chatkit` object.
+    ///     - userId: The user identifier to fetch a token for. The provider will sign a valid
+    ///     token for any user identifier requested, without applying any kind of authentication.
     /// - Returns:a `TokenProvider` which fetches tokens from the Chatkit Test Token Provider.
-    public static func chatkitTestTokenProvider(instanceLocator: String, userId: String) -> TokenProvider {
+    public static func createChatkitTestTokenProvider(instanceLocator: String, userId: String) -> TokenProvider {
         return CachingTokenProvider(
             delegateProvider: RetryingTokenProvider(
-                delegateProvider: testTokenHTTPCallProvider(
+                delegateProvider: createTestTokenHTTPCallProvider(
                     instanceLocator: instanceLocator, userId: userId
                 ),
                 retryAttempts: 3,
@@ -44,16 +50,22 @@ public struct ChatkitTokenProviders {
         )
     }
 
-    /// Create a `TokenProvider` which fetches tokens from an HTTP endpoint in your backend, with caching and retry implemented. This is the recommended implementation for production use.
+    /// Create a `TokenProvider` which fetches tokens from an HTTP endpoint in your backend, with
+    /// caching and retry implemented. This is the recommended implementation for production use.
     ///
     /// - Parameters:
     ///     - method: The HTTP method to use in the call, e.g. `GET`, `POST`
     ///     - host: The host name to be called, e.g. `example.com`
     ///     - path: The path component of the URL to be called, e.g. `/tokens`
-    ///     - headers: An optional map of headers to include in the request. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
-    ///     - queryParams: An optional map of query parameters to include in the request URL. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
-    /// - Returns: a `TokenProvider` which fetches tokens from the specified HTTP endpoint, and is wrapped in default caching and retry logic.
-    public static func httpTokenProvider(method: String, host: String, path: String, headers: [String: String]?, queryParams: [String: String]?) -> TokenProvider {
+    ///     - headers: An optional map of headers to include in the request. Here you can supply
+    ///     the session or other credientials which your endpoint might require to authenticate the
+    ///     request.
+    ///     - queryParams: An optional map of query parameters to include in the request URL. Here
+    ///     you can supply the session or other credientials which your endpoint might require to
+    ///     authenticate the request.
+    /// - Returns: a `TokenProvider` which fetches tokens from the specified HTTP endpoint, and is
+    /// wrapped in default caching and retry logic.
+    public static func createHTTPSTokenProvider(method: String, host: String, path: String, headers: [String: String]?, queryParams: [String: String]?) -> TokenProvider {
         return CachingTokenProvider(
             delegateProvider: RetryingTokenProvider(
                 delegateProvider: HTTPCallTokenProvider(
@@ -69,7 +81,7 @@ public struct ChatkitTokenProviders {
         )
     }
 
-    internal static func testTokenHTTPCallProvider(instanceLocator: String, userId: String) -> TokenProvider {
+    internal static func createTestTokenHTTPCallProvider(instanceLocator: String, userId: String) -> TokenProvider {
         // TODO: Implement:
         // extract host and instanceId from instanceLocator
         let host = "us1.pusherplatform.io"
--- a/Chatkit/Auth/TokenProvider.swift
+++ b/Chatkit/Auth/TokenProvider.swift
@@ -83,8 +83,8 @@ public class DefaultTokenProvider: TokenProvider {
     ///     the session or other credientials which your endpoint might require to authenticate
     ///     the request.
     ///     - queryParams: An optional map of query parameters to include in the request URL.
-    ///     Here you can supply the session or other credientials which your endpoint might
-    ///     require to authenticate the request.
+    ///     Here you can supply any query parameters which your endpoint might require to process
+    ///     the request.
     public convenience init(method: String,
                             host: String,
                             path: String,
@@ -109,8 +109,8 @@ public class DefaultTokenProvider: TokenProvider {
     ///     to be made, so that you can supply headers describing the current application user
     ///     session.
     ///     - queryParams: An optional async function which will be invoked when a request is about
-    ///     to be made, so that you can supply query parameters describing the current application
-    ///     user session.
+    ///     to be made, so that you can supply query parameters which your backend might require
+    ///     to process the request.
     public init(method: String,
                 host: String,
                 path: String,
--- a/Chatkit/Auth/TokenProvider.swift
+++ b/Chatkit/Auth/TokenProvider.swift
@@ -1,10 +1,3 @@
-//
-//  TokenProvider.swift
-//  Chatkit
-//
-//  Created by Mike Pye on 06/12/2019.
-//  Copyright © 2019 Pusher Ltd. All rights reserved.
-//
 // STUB FILE. This file demonstrates the desired interface, but contains to implementation
 // TODO: Move many of these types to PusherPlatform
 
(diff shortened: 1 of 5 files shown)


Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+extension String: Error { }
+
+// Re-export so that PusherPlatform import is not required in client code
+public typealias TokenProvider = PPTokenProvider
+public typealias TokenProviderResult = PPTokenProviderResult
+
+/// A factory for `TokenProvider`s.
+///
+/// The methods on this object create token provider instances which can interface with an HTTP endpoint in your backend, or the Chatkit Test Token Provider (for use during development)
+public struct ChatkitTokenProviders {
```
grzesiekko: Why is this prefixed with `Chatkit`? This should be derived from the module name.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+            host: host,
+            path: path,
+            headers: nil,
+            queryParams: [ "user_id": userId ]
+        )
+    }
+}
+
+/// CachingTokenProvider delegates the fetching of tokens to an underlying `delegateProvider`, retrying on failure
+public class CachingTokenProvider: TokenProvider {
```
grzesiekko: Perhaps it would be worth moving `CachingTokenProvider`, `RetryingTokenProvider` and `HTTPCallTokenProvider` classes to separate files?

marekoid:
Which parts/files shall be specifically reviewed as over 6K changed lines is a bit too much? ;)

grzesiekko:
> Which parts/files shall be specifically reviewed as over 6K changed lines is a bit too much? ;)

Only `Chatkit/Auth/TokenProvider.swift` and `Demo/Controllers/RoomListViewController.swift` contain meaningful changes.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+import PusherPlatform
+
+// TODO: Temporary
+extension String: Error { }
+
+// Re-export so that PusherPlatform import is not required in client code
+public typealias TokenProvider = PPTokenProvider
+public typealias TokenProviderResult = PPTokenProviderResult
+
+/// A factory for `TokenProvider`s.
```
marekoid: If first in the docs we explain that `ChatkitTokenProviders` is a factory wouldn't it be better to call it `ChatkitTokenProviderFactory`? Now it may sound a little bit more like it is a collection of ready to use (already instantiated) providers.

Also shouldn't `TokenProvider` be consistently called `ChatkitTokenProvider`? Also to better convey the information and avoid potential name clashes.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+///
+/// The methods on this object create token provider instances which can interface with an HTTP endpoint in your backend, or the Chatkit Test Token Provider (for use during development)
```
marekoid: The line length should no more that a half of the current length. It's not fitting GH view port and even if you have a massive screen it's hard for eyes to follow such long line and quickly jump to the next line (without readjusting or the risk to miss a line).

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+
+import Foundation
+import PusherPlatform
+
+// TODO: Temporary
+extension String: Error { }
+
+// Re-export so that PusherPlatform import is not required in client code
+public typealias TokenProvider = PPTokenProvider
+public typealias TokenProviderResult = PPTokenProviderResult
```
marekoid: Improving the platform lib is not in the scope? Wouldn't it be better to give them better names at the source? `P` prefix is not very clear. I think that `Chatkit` prefix will be the best or `Pusher` if it's something not only specific to `Chatkit`.

Regarding re-exporting couldn't it be handled here like `api` dependency in Gradle?

mdpye: Renaming everything is the platform library not to have a `PP` prefix (which is an Obj-C convention we want to drop) is a possibility, but not something I want to dive in to right now.
Improvements are in scope though.

The re-export is because swift imports work quite differently to android - they aren't automatic, and they are whole module, not per-type. So I don't want users of the SDK to have to `import PusherPlatform` at all to use the SDK.

Chatkit/Auth/TokenProvider.swift
mdpye: I'm not averse to the name, but lots of people have a very unreasonable reaction to the word "factory" :/

Chatkit/Auth/TokenProvider.swift
mdpye: This is a problem I'm having with XCode. It renders doc comments in variable width font, making it almost impossible to judge reasonable line lengths, and it doesn't auto-wrap at a reasonable length either.

I'll dig in the settings and see if I can have something reasonable enforced.

Chatkit/Auth/TokenProvider.swift
mdpye: No good reason. Should we change the module name to match (as Marek suggests), or drop the prefix from the things in here?

Chatkit/Auth/TokenProvider.swift
mdpye: Sure, if we decide to keep them. I don't know if there's a better way to introduce the desired common functionality?

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+    /// Create a `TokenProvider` which fetches tokens from an HTTP endpoint in your backend, with caching and retry implemented. This is the recommended implementation for production use.
+    ///
+    /// - Parameters:
+    ///     - method: The HTTP method to use in the call, e.g. `GET`, `POST`
+    ///     - host: The host name to be called, e.g. `example.com`
+    ///     - path: The path component of the URL to be called, e.g. `/tokens`
+    ///     - headers: An optional map of headers to include in the request. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
+    ///     - queryParams: An optional map of query parameters to include in the request URL. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
+    /// - Returns: a `TokenProvider` which fetches tokens from the specified HTTP endpoint, and is wrapped in default caching and retry logic.
+    public static func httpTokenProvider(method: String, host: String, path: String, headers: [String: String]?, queryParams: [String: String]?) -> TokenProvider {
```
marekoid: Wouldn't it be good to have `POST` as default? It's about creating an access session so probably POST fits it better than GET?

I reckon HTTPs will be used for the communication. Wouldn't it be worth to convey it at least in the docs? Now someone may be worried that it's not encrypted HTTP.

Chatkit/Auth/TokenProvider.swift
marekoid: Is the `struct` even required here if there are only `static func`? I have a very limited knowledge in Swift and may be asking silly questions, coming from the Kotlin world ;)

Chatkit/Auth/TokenProvider.swift
grzesiekko: 2-3 letter prefixes are Objective-C way of introducing a fake namespace (2 letters for Apple frameworks, 3 letters for third parties). Swift creates namespaces using modules defied in a `modulemap` file. I am not sure if that is a concern but it is possible for us to provide different names for Objective-C and Swift.

It is possible in Swift (and Objective-C) to import a single type:

```
import class PusherPlatform.Instance
```

Chatkit/Auth/TokenProvider.swift
grzesiekko: Currently the module is defined as `PusherChatkit`. What is the suggested new name?

When it comes to the token provider, at least in Swift, it would feel more natural to call the struct `TokenProvider`.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+///
+/// This class will make an HTTP request for every call to `fetchToken`, and in almost all cases it should be wrapped with caching and retry logic. See `httpTokenProvider(...)` for a factory method which returns a wrapped instance of this class.
+///
+/// - Parameters:
+///     - method: The HTTP method to use in the call, e.g. `GET`, `POST`
+///     - host: The host name to be called, e.g. `example.com`
+///     - path: The path component of the URL to be called, e.g. `/tokens`
+///     - headers: An optional map of headers to include in the request. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
+///     - queryParams: An optional map of query parameters to include in the request URL. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
+public class HTTPCallTokenProvider: TokenProvider {
```
marekoid: Why do we want to expose this type? Having it as an implementation detail behind the factory method wouldn't be better? I thought that's the motivation for the factory method.

Chatkit/Auth/TokenProvider.swift
grzesiekko: @marekoid It can be a different data structure as well, but you need to have something. `struct` is very lightweight and is probably a good choice from Swift perspective.

marekoid: Oh then it sounds like there are no top-level functions like in Kotlin ;)

grzesiekko: You can use top-level/global functions, but that is rarely used. Very often `static` or `class` variables\methods are wrapped in structs\classes to provide some more context.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+    ///
+    /// The Chatkit Test Token Provider is a token provider endpoint hosted by Pusher which returns tokens for the requested `userId` without authenticating the user. It is intended for development convenience and *must not be enabled in instances which host production data*.
+    ///
+    /// The test token provider must be enabled for your instance in the Chatkit dashboard.
+    ///
+    /// - Parameters:
+    ///     - instanceLocator: The locator string for your instance, which you also provide when initialising the `Chatkit` object.
+    ///     - userId: The user identifier to fetch a token for. The provider will sign a valid token for any user identifier requested, without applying any kind of authentication.
+    /// - Returns:a `TokenProvider` which fetches tokens from the Chatkit Test Token Provider.
+    public static func chatkitTestTokenProvider(instanceLocator: String, userId: String) -> TokenProvider {
```
marekoid: Wouldn't it be better to have `create` prefix so that it's clear it's a factory method and each call will create a new instance?

Chatkit/Auth/TokenProvider.swift
marekoid: Here function name provides full context.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
+//
+//  TokenProvider.swift
+//  Chatkit
+//
+//  Created by Mike Pye on 06/12/2019.
```
marekoid: What's the motivation for having "created by" comment? It feels like it's being used less and less often in the industry. After all a better info can be found in VCS.

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+/// HTTPCallTokenProvider makes calls to a specified HTTP endpoint and expects to receive a token from it.
+///
+/// This class will make an HTTP request for every call to `fetchToken`, and in almost all cases it should be wrapped with caching and retry logic. See `httpTokenProvider(...)` for a factory method which returns a wrapped instance of this class.
+///
+/// - Parameters:
+///     - method: The HTTP method to use in the call, e.g. `GET`, `POST`
+///     - host: The host name to be called, e.g. `example.com`
+///     - path: The path component of the URL to be called, e.g. `/tokens`
+///     - headers: An optional map of headers to include in the request. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
+///     - queryParams: An optional map of query parameters to include in the request URL. Here you can supply the session or other credientials which your endpoint might require to authenticate the request.
```
marekoid: Wouldn't it be better to just just suggest putting session id as a header? I know it's not about URLs for a browser, or an end user (and they won't be shared) but putting such info as query parameters is generally considered not a good practice. For example within the backend infrastructure there may be logging software which will be logging URLs.

Chatkit/Auth/TokenProvider.swift
mdpye: But there is no importing of anything smaller than an entire module, and no namespacing, so wrapping them in a struct is a way to offer some namespacing.

Chatkit/Auth/TokenProvider.swift
mdpye: Oh, that's cool. I haven't seen anyone use it!

Chatkit/Auth/TokenProvider.swift
mdpye: XCode adds it by default, I can delete it

Chatkit/Auth/TokenProvider.swift
mdpye: Header is definitely the better solution, but this is for making a request to an endpoint the customer is providing. We don't know how their auth works, or whether it's any good, but we can't really dictate requirements on it.

Chatkit/Auth/TokenProvider.swift
grzesiekko: Class cluster perhaps?
https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/ClassClusters/ClassClusters.html

Chatkit/Auth/TokenProvider.swift
marekoid: Still there won't be a clash on `createHttpTokenProvider` function name within our module.

Chatkit/Auth/TokenProvider.swift
mdpye: As a default for the `method` parameter? We could do, but I'm not sure it's worth much to the developer (who still has to check that method their backend exposes)

Chatkit/Auth/TokenProvider.swift
mdpye: For advanced usage (as yet undocumented), we want people to be able to make different implementations or vary different levels of the stack (network, retries, caching). The factory methods are just conveniences for the common use case, which is why these types are accessible

Chatkit/Auth/TokenProvider.swift
marekoid: They need to check but most likely it will be POST. Also they won't need to remember to type it in such case.

marekoid: What happens when client's session expires or is invalidated? Do they need to reinstantiate the whole Chatkit just to create a new TokenProvider? Will there'll be an option to set a new TokenProvider? If yes, probably it would be better to have `TokenProviderFactory` which takes the data that does not change as constructor params and then have a factory method like `createTokenProviderWithRequestHeader(name: String, value: String)` (excuse my Kotlin;) + other less convenient/popular factory methods for multiple headers or changing query params.

Chatkit/Auth/TokenProvider.swift
grzesiekko: That would be consistent with factory methods we already have in `Chatkit` class (entry point).

Chatkit/Auth/TokenProvider.swift
mdpye: When the token expires, the provider fetches a new one.

The most naive provider fetches a token on every call, which is why we wrap it in a `CachingTokenProvider` which re-uses a token until it expires.

marekoid: That means that the SDK gets 401 from the BE and calls fetch token on the token provider?

marekoid: I've checked and it looks like currently fetchToken is called on every request. Still my question above stays valid on what shall happen when token expires or is invalidated on the backend side but the local token provider is not aware of it. 

Still if fetch doesn't need to mean actual fetch, wouldn't it be better to name it as retrieval function, e.g. `getToken`?

Chatkit/Auth/TokenProvider.swift
mdpye: https://swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage

The official Swift style guide says they should all be called `make...`

grzesiekko: Should we rename `Chatkit` methods in that case?

mdpye: I'll make a separate PR to change them all...

marekoid: Doesn't feel like matching mental model of an Android dev so probably best to have it as `create` there.

Chatkit/Auth/TokenProvider.swift
marekoid: As long as we won't have factories of factories or more factories that actual types we shall be safe ;)

Chatkit/Auth/TokenProvider.swift
mdpye: > Still my question above stays valid on what shall happen when token expires or is invalidated on the backend side but the local token provider is not aware of it.

Yes, if the token is rejected by the backend, then a new one should be requested.

> Still if fetch doesn't need to mean actual fetch, wouldn't it be better to name it as retrieval function, e.g. getToken?

Aside: The convention on iOS is that the `get` prefix actually denotes expensive actions, not property access.

But to answer your question, it's always a fetch, but sometimes it's cached. It's always asynchronous, and it could hit the network for any given call, it's just that a lot of the time it's been optimised away. To me that means the name should reflect the "worst case".

Chatkit/Auth/TokenProvider.swift
mdpye: Yes, we will follow the appropriate conventions for each ecosystem.

mdpye:
@luismfonseca pointed out https://github.com/pusher/chatkit-client-js/issues/174, which is an interesting piece of feedback.

Our tokens are not the only ones to expire - the session headers we accept when constructing our token provider may themselves expire and need to be asynchronously provided!

marekoid:
> 
> 
> @luismfonseca pointed out [pusher/chatkit-client-js#174](https://github.com/pusher/chatkit-client-js/issues/174), which is an interesting piece of feedback.
> 
> Our tokens are not the only ones to expire - the session headers we accept when constructing our token provider may themselves expire and need to be asynchronously provided!

That's what I meant [here](https://github.com/pusher/chatkit-swift/pull/228#discussion_r355502619). Though I think we should keep our interface simple and accepting an async behaviour for that on our side would be going down a too complicated and too specific path.

marekoid:
> 
> 
> > @luismfonseca pointed out [pusher/chatkit-client-js#174](https://github.com/pusher/chatkit-client-js/issues/174), which is an interesting piece of feedback.
> > Our tokens are not the only ones to expire - the session headers we accept when constructing our token provider may themselves expire and need to be asynchronously provided!
> 
> That's what I meant [here](https://github.com/pusher/chatkit-swift/pull/228#discussion_r355502619). Though I think we should keep our interface simple and accepting an async behaviour for that on our side would be going down a too complicated and too specific path.

Though consistent with TokenProvider interface, for which token retrieval function is async. However here we would need to have that probably as well for providing query parameters so not sure if it won't be creating too complex interface for the clients.

It feels like we should optimise the API design for simplicity and correctness of the most popular solution. That is accepting an async fun for retrieving a single auth header (which may be expiring). Here also such auth header provider should have another function to let it know that their token provider rejected the request (due to expiry/invalidation of the auth header) so that the auth header provider can refresh it when we retry calling their retrieval function again.

All other cases can be always done by implementing own custom TokenProvider.

marekoid:
Still an overarching thought is that providing any generic implementations of TokenProvider is not trivial and it feels like it would be best to tackle it in a subsequent iteration.

grzesiekko:
> > > @luismfonseca pointed out [pusher/chatkit-client-js#174](https://github.com/pusher/chatkit-client-js/issues/174), which is an interesting piece of feedback.
> > > Our tokens are not the only ones to expire - the session headers we accept when constructing our token provider may themselves expire and need to be asynchronously provided!
> > 
> > 
> > That's what I meant [here](https://github.com/pusher/chatkit-swift/pull/228#discussion_r355502619). Though I think we should keep our interface simple and accepting an async behaviour for that on our side would be going down a too complicated and too specific path.
> 
> Though consistent with TokenProvider interface, for which token retrieval function is async. However here we would need to have that probably as well for providing query parameters so not sure if it won't be creating too complex interface for the clients.
> 
> It feels like we should optimise the API design for simplicity and correctness of the most popular solution. That is accepting an async fun for retrieving a single auth header (which may be expiring). Here also such auth header provider should have another function to let it know that their token provider rejected the request (due to expiry/invalidation of the auth header) so that the auth header provider can refresh it when we retry calling their retrieval function again.
> 
> All other cases can be always done by implementing own custom TokenProvider.

I agree that we should optimise the API design for simplicity and correctness. It feels very weird for me as an iOS developer to have to implement a token provider as a first step in integrating a third party SDK. I struggle to find an example of a third party SDK thet would require developers to do that. We are in control of the SDK and the backend. What is so unique in messaging that requires developers to go through that additional step (compared to other third part SDKs related to analytics, crash reporting, network traffic control, remote databases, etc.)?

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,137 @@
...
+            headers: nil,
+            queryParams: [ "user_id": userId ]
+        )
+    }
+}
+
+/// CachingTokenProvider delegates the fetching of tokens to an underlying `delegateProvider`, retrying on failure
+public class CachingTokenProvider: TokenProvider {
+    private let delegate: TokenProvider
+    private var token: String?
```
marekoid: Not sure about iOS but on Android it would be better to represent it as `CharSequence` so that it can be cleared (not just marked as free mem by GC). It's a usual recommended secure coding practice for such secrets.

grzesiekko: No GC on iOS. With ARC you can decide when it should be released.

Chatkit/Auth/TokenProvider.swift
marekoid: That's fine, on Android get with async interface will also convey that the operation may take longer. I think that we should abstract from calling it "fetch" if it's not always fetch. Indeed if we suggest caching and avoiding additional RTT hitting a separate host before every communication with Chatkit...

Alternative names could be: `provideToken` or `retrieveToken`. They also abstract from if it's local memory access or a fetch taking longer time. However simple `getToken` feels like the best and the most common name.

Now it may be understood that TokenProvider interface function is called "fetch" as we do caching outside of it. Some integrators may be surprised to see much higher hit rates than they expected for their token provider endpoint.

Chatkit/Auth/TokenProvider.swift
marekoid: How about clearing? Releasing a memory is one thing, clearing it with zeros or random data is another thing. Aren't Strings in iOS immutable? It is the case for Objective-C AFAIR.

Chatkit/Auth/TokenProvider.swift
marekoid: I understand I have just suggested not suggesting it in the docs for query params ;)

Chatkit/Auth/TokenProvider.swift
grzesiekko: This is an issue that has been never addressed by Apple. There are open radars (bug reports) regarding that. The problem is that this concerns only jailbroken devices which are not officially supported by Apple, so the priority of this issue is low.

Chatkit/Auth/TokenProvider.swift
marekoid: Still they could copy internal implementation (it's open source) and use it as a starting point for implementing their custom TokenProvider. Designing for extension is hard and establishes very intimate API contract ("is" as opposed to "uses"). Indeed already designing a good well established API just for instantiating those generic implementations is not trivial. I think we should keep the API surface minimal, only for the bits we're confident about and happy to commit to maintain as part of the API contract.

grzesiekko:
Is authenticating users while offline something that concerns us right now?

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,139 @@
...
+    public convenience init(method: String,
+                            host: String,
+                            path: String,
+                            headers: [String: String]?,
+                            queryParams: ([String: String]?)) {
+        self.init(method: method,
+                  host: host,
+                  path: path,
+                  getHeaders: HTTPSTokenProvider.wrapCredentialsValue(headers),
+                  getQueryParams: HTTPSTokenProvider.wrapCredentialsValue(headers))
```
marekoid:
```suggestion
                  getQueryParams: HTTPSTokenProvider.wrapCredentialsValue(queryParams))
```

Demo/Controllers/RoomListViewController.swift
```diff
@@ -64,8 +64,12 @@ class RoomListViewController: UITableViewController {
...
     
     private func connect() {
-        guard let chatkit = try? Chatkit(instanceLocator: "test:Instance:Locator") else {
-            return
+        let tokenProvider = ChatkitTestTokenProvider(instanceLocator: "test:Instance:Locator",
+                                                     userId: "olivia")
+        
+        guard let chatkit = try? Chatkit(instanceLocator: "test:Instance:Locator",
+                                         tokenProvider: tokenProvider) else {
+                                            return
```
mdpye: @grzesiekko 😨

Chatkit/Auth/TokenProvider.swift
marekoid: Regarding the TokenProvider interface I would like to also propose better naming of `clearToken` function: `refreshToken` which would indicate that the next call to `getToken` should return a fresh token. That is actually the trigger to call `clearToken`. It's being called when we get 401 passed down by our backend. Also currently having the token param there seems to be unnecessary complexity.

`clearToken` feels more like clearing it for security purposes when one stops using Chatkit. It feels even more weird when called in the context of an immutable String (which cannot be cleared) and moreover having another instance of immutable sting with that token content passed as a param. If we want to expose TokenProvider implementation which does caching then it should have a function `clearCachedToken` but it should not be a part of the generic TokenProvider interface.

It appears thought that iOS version of TokenProvider does not have `clearToken` function, at least I don't see it implemented in `HTTPSTokenProvider`. Is there a difference between declaration of TokenProvider on Android and iOS in the platform lib?

mdpye:
> Is authenticating users while offline something that concerns us right now?

No, not yet. And the earliest examples of offline support will involve queuing operations to do *when we get back online* and tokens are available again.

Initialising the SDK while offline will be a few iterations in.

mdpye:
> Here also such auth header provider should have another function to let it know that their token provider rejected the request (due to expiry/invalidation of the auth header) so that the auth header provider can refresh it when we retry calling their retrieval function again.

This is a good point, which could simplify things.

What do you think of the current suggestion, where the async providing is optional?

Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,139 @@
...
+    ///     to be made, so that you can supply headers describing the current application user
+    ///     session.
+    ///     - queryParams: An optional async function which will be invoked when a request is about
+    ///     to be made, so that you can supply query parameters describing the current application
+    ///     user session.
+    public init(method: String,
+                host: String,
+                path: String,
+                getHeaders: AsyncCredentialsCall?,
+                getQueryParams: AsyncCredentialsCall?) {
```
marekoid: Wouldn't it be worth to call it more generic terms, e.g. "KeyValue" instead of "Credentials"? After all it could be that there's an auth header that can expire or get invalidated but query params still are needed and have nothing to do with auth (probably they're static too but that's another issue with trying to aim for such generic implementation).

Still not sure how we can fit in notifying about the need to refresh of the auth token. Still leaning [towards](https://github.com/pusher/chatkit-swift/pull/228#issuecomment-563986272) the most.

Chatkit/Auth/TokenProvider.swift
mdpye: This has likely changed around a bit since I simplified and made the caching and retry an internal concern. Perhaps we can talk about this in person.

mdpye:
Merged as #232

PR #231, opened 2019-12-10
Default Xcode formatting for Swift

grzesiekko:
### What?

- Default source code formatting for Swift (Ctrl-I in Xcode).

### Why?

- Speed up development
- Get rid of empty changes created by Xcode
- Readability of source code

----

### DIFF
--- a/Chatkit/Chatkit/Chatkit.swift
+++ b/Chatkit/Chatkit/Chatkit.swift
@@ -115,7 +115,7 @@ public class Chatkit {
     }
     
     // MARK: - Constructing paged data providers
-
+    
     /// Creates an instance of `UsersProvider`.
     ///
     /// - Parameters:
@@ -153,7 +153,7 @@ public class Chatkit {
     }
     
     // MARK: - Constructing real time data providers
-
+    
     /// Creates an instance of `JoinedRoomsProvider`.
     ///
     /// This will provide access to a real time set of `Room`s that the current user is a member of.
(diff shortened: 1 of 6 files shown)


mdpye:
While I agree with the idea of auto-formatting the code, I really struggle with the method parameter line breaking style in here, vs the one it is replacing (which is the Google style: https://google.github.io/swift/#line-wrapping)

It:
- Is very difficult to read (accepted, might be subjective)
- Creates whitespace changes across many lines if method names change
- (most importantly) Doesn't actually claim much of the following line for use, making it difficult to break lines to the 100 chars we agreed.
  - This compounds as you move to the right, as each new indentation is the width of a method name, not a fixed indent.

Can XCode be asked to indent "Google style"?

grzesiekko:
> While I agree with the idea of auto-formatting the code, I really struggle with the method parameter line breaking style in here, vs the one it is replacing (which is the Google style: https://google.github.io/swift/#line-wrapping)
> 
> It:
> 
> * Is very difficult to read (accepted, might be subjective)
> * Creates whitespace changes across many lines if method names change
> * (most importantly) Doesn't actually claim much of the following line for use, making it difficult to break lines to the 100 chars we agreed.
>   
>   * This compounds as you move to the right, as each new indentation is the width of a method name, not a fixed indent.
> 
> Can XCode be asked to indent "Google style"?

No one uses Google style apart from Google.

mdpye:
I can't get it to apply to some code I've just written. Ctrl-I is re-indent, it doesn't change the line breaking for me.

PR #232, opened 2019-12-10
Auth interface v2

mdpye:
This is a fresh PR incorporating most of the feedback from #228, because it grew very large, and I wanted to squash before merge.

Any further comments?

### DIFF
--- a/Chatkit/Auth/TokenProvider.swift
+++ b/Chatkit/Auth/TokenProvider.swift
@@ -1,0 +1,139 @@
+// STUB FILE. This file demonstrates the desired interface, but contains to implementation
+// TODO: Move many of these types to PusherPlatform
+
+import Foundation
+import PusherPlatform
+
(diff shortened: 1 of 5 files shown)


Chatkit/Auth/TokenProvider.swift
```diff
@@ -0,0 +1,139 @@
...
+    
+    public func fetchToken(completionHandler: @escaping (PPTokenProviderResult) -> Void) {
+        self.delegate.fetchToken(completionHandler: completionHandler)
+    }
+    
+}
+
+public typealias AsyncKeyValueCall = (CredentialsCompletionHandler) -> ()
+
+public typealias CredentialsCompletionHandler = (KeyValueResult) -> ()
```
marekoid: It looks like "Credentials" here managed to hide itself from the renaming ;)

PR #233, opened 2019-12-10
Update the beams sdk which supports multiple instances and consume

daniellevass:
### What?
We updated the beams sdk and updated where we use it to support multiple instances.

To test this we added a chatkit instance id into the demo app + instantiated beams e.g. our app delegate looked like the following:

```swift
 let instanceLocator = "CHATKIT_INSTANCE_LOCATOR"

        let pusherChat = ChatManager(
            instanceLocator: instanceLocator,
            tokenProvider: PCTokenProvider(url: "CHATKIT_TOKEN_PROVIDER"),
            userID: "danielle",
            logger: TestLogger()
        )

        self.pusherChat = pusherChat

        // Initiate the registration process with Apple Push Notification service.
        
        ChatManager.registerForRemoteNotifications()
        
        self.beamsClient.start(instanceId: "BEAMS_INSTANCE_ID")
        self.beamsClient.registerForRemoteNotifications()
```

### Why?
This will enable people to use beams + chatkit in the same app

----

### DIFF
--- a/Sources/ChatManager.swift
+++ b/Sources/ChatManager.swift
@@ -401,8 +401,6 @@ func pathFriendlyVersion(of component: String) -> String {
 
 #if os(iOS) || os(macOS)
 // MARK: Beams
-private let pushNotifications: PushNotifications = PushNotifications.shared
-
 extension ChatManager {
     /**
      Register device token with PushNotifications service.
@@ -410,7 +408,7 @@ extension ChatManager {
      - Parameter deviceToken: A token that identifies the device to APNs.
      */
     public static func registerDeviceToken(_ deviceToken: Data) {
-        pushNotifications.registerDeviceToken(deviceToken)
+        PushNotifications.shared.registerDeviceToken(deviceToken)
     }
     /**
      Register to receive remote notifications via Apple Push Notification service.
@@ -420,7 +418,7 @@ extension ChatManager {
      - SeeAlso:  `registerForRemoteNotifications(options:)`
      */
     public static func registerForRemoteNotifications() {
-        pushNotifications.registerForRemoteNotifications()
+        PushNotifications.shared.registerForRemoteNotifications()
     }
     #if os(iOS)
     /**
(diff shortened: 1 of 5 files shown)


Sources/ChatManager.swift
```diff
@@ -401,16 +401,14 @@ func pathFriendlyVersion(of component: String) -> String {
...
-
 extension ChatManager {
     /**
      Register device token with PushNotifications service.
 
      - Parameter deviceToken: A token that identifies the device to APNs.
      */
     public static func registerDeviceToken(_ deviceToken: Data) {
-        pushNotifications.registerDeviceToken(deviceToken)
+        PushNotifications.shared.registerDeviceToken(deviceToken)
```
mdpye: I thought we were not going to use the `shared` instance in Chatkit, so that it's available for customers in their own usage?

mdpye: That certainly appears to be how it was done in Android.

Sources/PCCurrentUser.swift
```diff
@@ -1285,27 +1285,24 @@ public typealias PCRoomsCompletionHandler = ([PCRoom]?, Error?) -> Void
...
 extension PCCurrentUser {
     /**
      Start PushNotifications service.
      */
     public func enablePushNotifications() {
-        pushNotifications.start(instanceId: self.v6Instance.id)
-        self.setUser(self.id)
-        ChatManager.registerForRemoteNotifications()
-    }
+        let beamsClient = PushNotifications.init(instanceId: self.v6Instance.id)
```
mdpye: Is this calling the `PushNotifications` constructor, or is there a method called `init`?

Sources/ChatManager.swift
luismfonseca: You're correct that the intention is that one. On the swift SDK, this function will apply to all instances that are in a started state, so we can use the singleton here. Likewise for the other function, `registerForRemoteNotifications`, which just prompts the UI for the user to accept the push notifications permission.

mdpye: Shouldn't the functions be `static` then?

If not, then I think a comment is definitely in order.

Sources/PCCurrentUser.swift
daniellevass: .init is just explicitly calling the initialiser, we have removed this word (and it still does the same thing)

Sources/ChatManager.swift
daniellevass: They'd both do the same thing. But I see your point, I've updated the PR to use `PushNotificationsStatic` instead of `PushNotifications.sahred` to make it clearer what's happening.

luismfonseca: `PushNotifications.shared` was left as a backwards compatibility, but it just points to the static version. But there's no reason why we can't be explicit here.

PR #246, opened 2020-01-23
Test Token Provider

grzesiekko:
### What?

- New `TestTokenProvider` that helps in connecting with Pusher's test token service.
- Add `System Tests` target

### Why?

Chatkit 2.0 roadmap.

----


System Tests/Environment/Environment.swift
```diff
@@ -0,0 +1,20 @@
...
+    // MARK: - Private methods
+    
+    private static func variable(named name: String, ciValue: String) -> String {
+        guard let value = ProcessInfo.processInfo.environment[name], value != ProcessInfo.unspecifiedValue else {
+            assert(!ciValue.hasPrefix("$(") && !ciValue.hasSuffix(")"), "Please specify the value of the variable in either the enviroment variables section of your current run scheme or inject the value using Swift Variable Injector from Bitrise.")
+            return ciValue
+        }
+        
+        return value
+    }
```
OliverPearmain-Pusher: Its not straight forward to ascertain what exactly is going on here.   I _think_ the codes been update so that its now possible to define your instance locator either in the scheme or in the `CI.swift` file, if thats the case perhaps you could add a couple of additional methods to make it more obvious like...
`private static environmentVariable(fromSchemeNamed: ) -> String?`
and 
`private static environmentVariable(fromCISwiftFile: ) -> String?`

System Tests/Environment/CI.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+struct CI {
```
OliverPearmain-Pusher: I had anticipated this would be a text file in the *root* of the repo so that it can be consumed by the demo app and all test targets as opposed to a source file.  We can fix this later though if you prefer.

Demo/Environment/Environment.swift
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// MARK: - Validation
+
+extension Environment {
+    
+    // MARK: - Internal methods
+    
+    static func validate() {
+        assert(Environment.instanceLocator != "<UNSPECIFIED>", "Please provide above your instance locator in order to run the app. You can find your instance locator Credentials tab of your Dashboard.")
```
OliverPearmain-Pusher: I it might be better if this behaved the same way as the tests, i.e. it checks for an environment variable and then a file.  Again this is something we can resolve later if you prefer.

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
```diff
@@ -0,0 +1,78 @@
...
+    private static let urlScheme = "https"
+    private static let urlHost = "pusherplatform.io"
+    private static let urlService = "services/chatkit_token_provider"
+    private static let urlResource = "token"
+    
+    /// The locator for your instance, the same value from the Dashboard which you use to construct
+    /// the Chatkit object.
+    public let instanceLocator: String
+    
+    /// The userID to fetch tokens for. A token will always be signed for this userID without any
```
OliverPearmain-Pusher: "The userID for whom to fetch tokens" maybe reads a little better I think (this is repeated below also)

Unit Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
```diff
@@ -0,0 +1,173 @@
+import XCTest
+import Mockingjay
+@testable import PusherChatkit
+
+class TestTokenProviderTests: XCTestCase {
+    
+    // MARK: - Properties
+    
+    var testURL: URL!
+    var matcher: (URLRequest) -> Bool = uri("https://instance.pusherplatform.io/services/chatkit_token_provider/test/locator/token?user_id=testUserID")
```
OliverPearmain-Pusher: I don't think you need to give this an initial value because you're always setting it in the `setUp` method.

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
```diff
@@ -0,0 +1,78 @@
...
+    
+    private static let userIdentifierQueryItemName = "user_id"
+    private static let urlScheme = "https"
+    private static let urlHost = "pusherplatform.io"
+    private static let urlService = "services/chatkit_token_provider"
+    private static let urlResource = "token"
+    
+    /// The locator for your instance, the same value from the Dashboard which you use to construct
+    /// the Chatkit object.
+    public let instanceLocator: String
```
OliverPearmain-Pusher: As discussed I think the `public` properties here should be `private`.

System Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+                
+            default:
+                XCTFail("Failed to retrieve token from the web service.")
+            }
+            
+            expectation.fulfill()
+        }
+        
+        waitForExpectations(timeout: 5.0)
+    }
```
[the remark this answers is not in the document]
grzesiekko: Sure, I can add such comments.

Unit Tests/Helpers/String+Extensions.swift
```diff
@@ -1,6 +1,6 @@
 import XCTest
 
-extension String {
+extension String: Error {
```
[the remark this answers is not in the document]
grzesiekko: I have an issue with this change as well. It's part of the PR only because the old draft of token providers defined `extension String: Error`. I deleted that and our unit tests stopped to compile due to the fact that `toData()` method throws a string. I think this should be improved, but I wasn't sure if that should be part of this PR.

Unit Tests/Helpers/DummyTokenProvider.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import PusherPlatform
+
+class DummyTokenProvider: TokenProvider {
```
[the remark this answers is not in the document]
grzesiekko: I will rename this class to `FakeTokenProvider`.

Demo/Controllers/RoomListViewController.swift
```diff
@@ -64,8 +64,9 @@ class RoomListViewController: UITableViewController {
     }
     
     private func connect() {
-        guard let chatkit = try? Chatkit(instanceLocator: "test:Instance:Locator", tokenProvider: TestTokenProvider()) else {
-            return
+        guard let tokenProvider = try? TestTokenProvider(instanceLocator: "", userID: "joe"),
```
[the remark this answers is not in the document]
grzesiekko: Sure, I can add that.

Chatkit/Networking/Instance Locator/InstanceLocator.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+struct InstanceLocator {
```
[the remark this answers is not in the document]
grzesiekko: I moved `Instance Locator` to `PusherPlatform`. There is a separate PR for that change in `PusherPlatform` repository.

System Tests/Environment/CI.swift
grzesiekko: I would structure this differently as well, but the only usable Bitrise extension (Swift Variable Injector) that allows us to inject custom configuration uses Swift files. You can find more details about the extension and its configuration here: https://medium.com/@lucianoalmeida1/continuous-integration-environment-variables-in-ios-projects-using-swift-f72e50176a91

System Tests/Environment/Environment.swift
grzesiekko: Sure, I can do that.

Demo/Environment/Environment.swift
grzesiekko: I was implemented like that initially, but than we discussed that our preference would be to modify a Swift file rather that set an environment variable in the run scheme.

I happy to change that back to environment variables.

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
```diff
@@ -0,0 +1,78 @@
...
+    
+    /// The userID to fetch tokens for. A token will always be signed for this userID without any
+    /// authentication being applied.
+    public let userID: String
```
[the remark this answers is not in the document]
grzesiekko: `userIdentifier` would be my preference as well, but `userID` was defined in the draft of the public interface. I am happy to change that. @mdpye

Unit Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
grzesiekko: I forgot to remove this initialization.

Unit Tests/Helpers/DummyTokenProvider.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: Perfect, thanks

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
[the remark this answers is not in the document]
mdpye: Yeah, go with "identifier", as used everywhere else in the SDK. It's a peculiarity of the iOS ecosystem, my fingers type "id" without we asking them to.

Although is android it's `userId`. Fun fact, in Go it's `userID` or the (non-configurable) linter will reject your code, because the authors don't understand the different between an abbreviation and an acronym...

By which rambling, I mean to say that any interface I propose ("specify") should be filtered for language bias, idiom and confusion :)

System Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: Awesome

Demo/Controllers/RoomListViewController.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: 👌  nice one

Chatkit/Networking/Instance Locator/InstanceLocator.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: Oh sweet, great work!

System Tests/Environment/CI.swift
OliverPearmain-Pusher: Aha, OK, I'll check those docs.  Thanks for sharing.

Demo/Environment/Environment.swift
OliverPearmain-Pusher: Leave it for now then.  👍   I'll check out those docs you linked and if we do come up with a better way for this to work then we can address it in the future.

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: I personally don't care whether we use `ID` `Id` or `Identifier`.  I just want to be consistent :-)

Unit Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
OliverPearmain-Pusher: 👀  :-)

Unit Tests/Helpers/String+Extensions.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: I think we possibily need to have a chat about what types of error we might expose in the public interface (and internally as well TBH).  This may change as a result, so leave it for now and it might change in the future 👍

System Tests/Environment/Environment.swift
OliverPearmain-Pusher: 👍  ta

OliverPearmain-Pusher: Ace, this is so much better :-)

Demo/Controllers/RoomListViewController.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: I think you added `TestUser.joe` but you forgot to consume it here :-)

System Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: This `testShouldRetrieveTokenFromTestTokenSerivce` test didn't get updated 😢

Unit Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
```diff
@@ -0,0 +1,224 @@
...
+        var result: AuthenticationResult?
+        
+        tokenProvider.fetchToken { authenticationResult in
+            result = authenticationResult
+            expectation.fulfill()
+        }
+        
+        /******************/
+        /*----- THEN -----*/
+        /******************/
```
OliverPearmain-Pusher: Thanks for updating all these tests with given, when & then 😃

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+import PusherPlatform
+
+/// TestTokenProvider retrieves tokens from the Chatkit service's Test Token Provider, which
+/// is for development user only, and must be enabled for your instance in the Chatkit Dashboard.
```
OliverPearmain-Pusher: Sorry didn't spot this earlier, should this be "development use" rather than "development use**r**"?

Demo/Controllers/RoomListViewController.swift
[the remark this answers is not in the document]
grzesiekko: We can discard this particular line for now. It is also missing instance locator. `RoomViewController` is not implemented at the moment.

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
```diff
@@ -0,0 +1,68 @@
...
+    private let nestedTokenProvider: DefaultTokenProvider
+    
+    // MARK: - Initializers
+    
+    /// Create an TestTokenProvider which retrieves tokens from the Chatkit service's Test Token Provider
+    ///
+    /// - Parameters:
+    ///     - instanceLocator: The locator for your instance, the same value from the Dashboard
+    ///     which you use to construct the Chatkit object.
+    ///     - userID: The userID for whom to fetch tokens. A token will always be signed for this userID
```
OliverPearmain-Pusher: Think the comment needs updating "userID" -> "userIdentifer"

Chatkit/Networking/Instance Locator/InstanceLocator.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: This class/file can now be deleted.

System Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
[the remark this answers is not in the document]
OliverPearmain-Pusher: 👍

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
OliverPearmain-Pusher: 👍

Chatkit/Authentication/Token Providers/TestTokenProvider.swift
OliverPearmain-Pusher: 👍

PR #251, opened 2020-01-27
Support for Swift Package Manager

grzesiekko:
### What?

- Add `Package.swift` in order to define `PusherChatkit` package.
- Integrate demo app with `PusherChatkit` using Swift Package Manager.

### Why?

2.0 roadmap + feature request.

----

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/pusher/push-notifications-swift.git",
         "state": {
           "branch": null,
-          "revision": "7f074a47beab45ca020806c06c88e65c80a7227e",
-          "version": "3.0.2"
+          "revision": "b2c163bab17a509edadd9b02449101a097fa4328",
+          "version": "3.0.3"
         }
       },
       {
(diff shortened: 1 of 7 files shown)


Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "PushNotifications",
+        "repositoryURL": "https://github.com/pusher/push-notifications-swift.git",
+        "state": {
+          "branch": null,
+          "revision": "7f074a47beab45ca020806c06c88e65c80a7227e",
+          "version": "3.0.2"
```
OliverPearmain-Pusher: Should we be on v 3.0.2 of push-notifications-swift?

OliverPearmain-Pusher: 👍

grzesiekko: This file is automatically generated by Xcode. It can be updated to `3.0.3` it doesn't have to as Xcode will resolve to the most recent version anyway (this file is only important for command line dependency resolution). `3.0.3` adds a fix for Objective-C which we don't support.

grzesiekko:
Closing due to a problem with GitHub/Bitrise setup

PR #254, opened 2020-01-27
Project Cleanup and Groups Reorg

OliverPearmain-Pusher:
### What?
Reorganised some of the project groups. Also removed some redundant classes/files.

### Why?
So its more obvious which files live where.

----


grzesiekko:
This is a matter of personal preference, but there are a few things that I would change in the new structure:

1. Keep nested groups (directories) on the top of every group.
2. Use spaces when naming groups (`ViewModels` -> `View Models`) due to consistency and an old Xcode convention for OS X projects. I personally don't think that we should ever use command line to rename things in our project.
3. It feels a bit weird to me that the entry point is in the same group as the token provider, types and connection status.
4. Introduction of the group `Public` leads to a situation in which `ConnectionStatus` is no longer placed in `Network` group (which feels like a natural place for this enumeration).

OliverPearmain-Pusher:
1. Are you saying that groups should srictly either contain only groups or only files but never a mix of files and groups?  If so, what is your motivation for this, other than personal preference?  I've never worked on any project, in any IDE, filesystem or langauge that stipulated this.   It seems to me that this rule would result in a) a deeper hierachary that necessary and b) an overhead of naming and maintaining a group that isn't really required.

2. Consistency is surely naming a group the same as whats inside it?  If the group contains a load of classes named "*ViewModel.swift" then surely you'd name the parent group "ViewModel" or "ViewModels" and nothing else.   I've never heard anything about an old Xcode convention and if that is the case it was probably long before the file system sync'ed with the group structure so is probably irrelevant now.

3. Could you elaboarate on why you find it weird?  A client is always going to use the entry point and the token provider in conjunction with each other.  To use the entry point you have to pass a token provider into it.  You'd never use one without the other.  So IMO they may as well live alongside each other in the project.

4. I think your point really highlights the issue I have with overuse of groups in general (which is exactly what this PR aims to address).  Its very difficult to classify files/classes in a project because there is always overlap between architectural bounaries and class dependencies.  So trying to classify everything into strict lower level groups is actually really hard and a pain.  It quickly becomes impossible to find anything, and tedious if you do know where things are because you have to click x? more twistys to get at the files.  Ultimately developers just end up using quick find to access things at which point you may as well have a flat structure anyway :-)

grzesiekko:
If you feel strongly about this please feel free to merge this PR without my approval.

OliverPearmain-Pusher:
I'm quite confident we can reach an agreement on this without stonewalling :-)

I've asked @mdpye  to cast his eye over the discussion to see if he has any preference but I think it will help if you might answer the questions that I posed above so that its possible to consider both sides of the argument please?  

It seems there are two questions to resolve...
1. Do we use spaces in group/directory names or not
2. Do we strictly try to group/categorise everthing or not

OliverPearmain-Pusher:
Closing PR as requested :-)

PR #257, opened 2020-01-29
Redux Store

OliverPearmain-Pusher:
### What?

- Added `Dependencies`, `Store`, `StoreBroadcaster` and `State` entities.
- Moved certain test extensions and shared test code into `Test Utilities`

### Why?
- So we can model the redux state and so work can commence on `Reducers`

----


Unit Tests/Tests/Networking/Model/Subscription/PresenceSubscription/Wire.Event.PresenceState+Decodable.Tests.swift
```diff
@@ -1,4 +1,5 @@
 import XCTest
+@testable import TestUtilities
```
grzesiekko: Does it make any sense to mark `TestUtilities` imports as `@testable`. `TestUtilities` should have its own public interface and we should not relay on internal implementations provided by it.

Chatkit/Redux/Store/State.swift
```diff
@@ -0,0 +1,34 @@
+
+enum Internal {}
```
grzesiekko: Isn't this a code smell that this fake namespace is not required and regular Swift access controls should be sufficient?

Chatkit/Redux/Store/State.swift
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+extension Internal.User: Equatable {}
+
+struct State {
+    let currentUser: Internal.User?
+    let joinedRooms: [Internal.Room]
+    
+    static var emptyState: State {
```
grzesiekko: It makes sense in my opinion to explicitly mark this as `let`.

It would also feel a bit more "Swifty" when named `empty` (`State.empty` instead of `State.emptyState`).

Test Utilities/Doubles/Dependencies.swift
```diff
@@ -0,0 +1,25 @@
+import XCTest
+@testable import PusherChatkit
+
+let DummyInstanceLocator = "dummy:instance:locator"
```
grzesiekko: There's already a test instance locator defined in an extension to `XCTest`: `Networking.testInstanceLocator`. It would be good to merge both into one.

Unit Tests/Tests/Redux/Store/ConcreteDependencyTests.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    func test_initWithInstanceLocator_success() {
+        
+        let dependencies = ConcreteDependencies(instanceLocator: DummyInstanceLocator)
+        
+        XCTAssertNotNil(dependencies.storeBroadcaster as? ConcreteStoreBroadcaster)
+        XCTAssertNotNil(dependencies.store as? ConcreteStore)
+    }
+    
+    func test_initWithInstanceLocatorAndOverride_success() {
```
grzesiekko: Where do we want to draw a line between tests following BDD style and tests that do not follow such style?

Chatkit/Redux/Store/State.swift
```diff
@@ -0,0 +1,34 @@
+
+enum Internal {}
+
```
grzesiekko: Where are unit tests that test `State`, `Room` and `User` in isolation?

Chatkit/Redux/Store/StoreBroadcaster.swift
```diff
@@ -0,0 +1,48 @@
...
+    
+    private var listeners = [StoreListener]()
+    
+    init(dependencies: Dependencies) {
+        self.dependencies = dependencies
+    }
+    
+    // MARK: StoreBroadcaster
+    
+    func register(_ listener: StoreListener) -> State {
```
grzesiekko: Why do we return `State` from this method?

Chatkit/Redux/Store/Store.swift
```diff
@@ -0,0 +1,110 @@
+
+enum Action {
+    case subscriptionEvent(Wire.Event.EventType)
+    case received(user: Wire.User)
+    case fetching(userWithIdentifier: String)
+}
+
+extension Action: Equatable {}
+
+protocol StoreDelegate: AnyObject {
```
grzesiekko: Is there any reason for this protocol to inherit from `AnyObject` instead of `class`?

Chatkit/Redux/Store/Store.swift
```diff
@@ -0,0 +1,110 @@
...
+        
+        let existingState = self.state
+        
+        let newState: State
+        
+        // TODO: this whole block needs to move into a `Reducer` or `ReductionManager`
+        // (and therefore this does not have full test coverage at present)
+        switch action {
+            
+        case let .subscriptionEvent(eventType):
```
grzesiekko: This might become hard to maintain as a single method. It's probably worth breaking this code into a few separate methods.

Chatkit/Redux/Store/Store.swift
```diff
@@ -0,0 +1,110 @@
...
+
+            case let .removedFromRoom(removedFromRoom):
+                // TODO:
+                newState = State(
+                    currentUser: existingState.currentUser,
+                    joinedRooms: existingState.joinedRooms.filter { $0.identifier != removedFromRoom.roomIdentifier }
+                )
+                
+            default:
+                fatalError()
```
grzesiekko: I would prefer to replace `fatalError()` with `preconditionFailure()` as the first puts system path of the source file into the compiled binary.

More details can be found here: https://lapcatsoftware.com/articles/fatalError.html

Chatkit.xcodeproj/project.pbxproj
```diff
@@ -1707,6 +1846,10 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(PLATFORM_DIR)/Developer/Library/Frameworks",
```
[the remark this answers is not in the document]
OliverPearmain-Pusher: Yes, exactly.

Unit Tests/Tests/Networking/Model/Subscription/PresenceSubscription/Wire.Event.PresenceState+Decodable.Tests.swift
OliverPearmain-Pusher: I asked myself the same question.  We can either explicitly mark things (well pretty everything I guess) as `public` in `TestUtilitilies` or just use `@testable`.   

If you'd like me to change to `public` then just say and I'll make it so.

grzesiekko: I think that `TestUtilities` is a framework that provides us with functionality that should have its own public interface. `@testable` might indicate a module that we are about to test.

Chatkit/Redux/Store/State.swift
OliverPearmain-Pusher: I need a namespace otherwise `Room` and `User` etc will clash with their `public` counterparts.  

At any rate this entire class is just a WIP and I expect it to be entirely rewritten.  This will become your baby since you are working on the `State` and `Reducers`.  You are free to arrange and name this as you wish. 

If you'd like me to change it to something else for this PR then make a suggestion and I'll do so.

Chatkit/Redux/Store/State.swift
OliverPearmain-Pusher: Both good suggestions 👍 .  Will change.

Chatkit/Redux/Store/State.swift
OliverPearmain-Pusher: All these classes are just stubbed (not a test sense) for you to flesh out.  They are effectively not implemented yet.  You will need to write these tests as part of your work on adding the  `State` and `Reducer`s.

Chatkit/Redux/Store/Store.swift
OliverPearmain-Pusher: `AnyObject` is preferred over `class` from Swift 4.1 onwards but I'm not fully aware of the reasoning.

Something to do with this...
https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md

Also this...
https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#anyObjectProtocol

Chatkit/Redux/Store/Store.swift
OliverPearmain-Pusher: See the comment a few lines above... 
```
        // TODO: this whole block needs to move into a `Reducer` or `ReductionManager`
        // (and therefore this does not have full test coverage at present)
```

It will undoubtly be split because each parsing block should end up in a dedicated `Reducer`.  This will form part of your work to implement `Reducers`.

Chatkit/Redux/Store/Store.swift
OliverPearmain-Pusher: These are only temporary but I'm happy to change and will try and remember to use `preconditionFailure()` going forward.  Thanks for the share.

Chatkit/Redux/Store/StoreBroadcaster.swift
OliverPearmain-Pusher: Otherwise the object thats registering has no way of knowing the state until the next time the state changes and the delegate method is invoked.

Unit Tests/Tests/Redux/Store/ConcreteDependencyTests.swift
OliverPearmain-Pusher: I am attempting for all unit tests to follow BDD style.  I haven't given these tests my full attention and I've spotted some cases that aren't tested.  I will address this now.

Test Utilities/Doubles/Dependencies.swift
OliverPearmain-Pusher: Agreed, we should only have one, I will attempt to merge.  I think we will need `valid` and `invalid` versions of the instance locators actually.

Unit Tests/Tests/Networking/Model/Subscription/PresenceSubscription/Wire.Event.PresenceState+Decodable.Tests.swift
OliverPearmain-Pusher: Completely agree, will change.

Chatkit/Redux/Store/Store.swift
OliverPearmain-Pusher: This now done FYI

Unit Tests/Tests/Redux/Store/ConcreteDependencyTests.swift
OliverPearmain-Pusher: I have now updated this test fixture entirely so I think its clear that all the tests are BDD and the test signatures make clear the "given" context.

Chatkit/Redux/Store/State.swift
OliverPearmain-Pusher: This now done FYI

Unit Tests/Tests/Networking/Model/Subscription/PresenceSubscription/Wire.Event.PresenceState+Decodable.Tests.swift
OliverPearmain-Pusher: This now done FYI

Test Utilities/Doubles/Dependencies.swift
OliverPearmain-Pusher: FYI this is now resolved.  Because I've switched all my stuff to depend on the `InstanceLocator` entity I don't ever need to use a string.  The string is now removed.

Chatkit/Redux/Store/State.swift
grzesiekko: Let's leave `Internal` then.

PR #258, opened 2020-01-29
OKP `InstanceLocator` file

OliverPearmain-Pusher:
### What?
Makes it possible for developers to define their instance locator in a file named `InstanceLocator` in the root of the repo.

### Why?
Developers would have to do this once and then the file/value remains even if changing branches or resetting your working copy and there’s no possibility of it being accidentally checked in (unlike with the env-var in the xcschemes at present).

----

### DIFF
--- a/Chatkit.xcodeproj/project.pbxproj
+++ b/Chatkit.xcodeproj/project.pbxproj
@@ -1919,16 +1919,12 @@
 		73A9394723E1FB800025AF90 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CODE_SIGN_STYLE = Automatic;
-				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = Debug;
 		};
 		73A9394823E1FB800025AF90 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CODE_SIGN_STYLE = Automatic;
-				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = Release;
 		};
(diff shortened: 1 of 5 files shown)


Chatkit.xcodeproj/project.pbxproj
```diff
@@ -1824,6 +1916,22 @@
 			};
 			name = Release;
 		};
+		73A9394723E1FB800025AF90 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				CODE_SIGN_STYLE = Automatic;
```
grzesiekko: Does this target require to be code signed?

OliverPearmain-Pusher: I shouldn't have thought so.  All the build settings are just the Xcode defaults for an aggregate target.  

I can try and align the build settings with the other targets if you'd like?

grzesiekko: I would personally aim to have as few build settings overrides as possible and keep most of the configuration of the project level.

Code signing has been disabled for all the other target because we lost access to our Apple Developer Program and this was causing issues on Bitrise.

OliverPearmain-Pusher: I completely agree with the sentiment.  Ideally we'd use xcconfig's but thats not something we need to implement anytime soon. 

I have removed *all* of the build settings for `MakeInstanceLocatorFile`. 

Just reaslised that I named the target without thinking.  I'm happy to rename `Make InstanceLocator File` for consistency if you'd prefer?

grzesiekko: Just keep it as it is.

PR #260, opened 2020-01-31
OKP - Update in line with PusherPlatform changes

OliverPearmain-Pusher:
### What?
The PusherPlatform `Token.token` property  was renamed to `Token.value`.  Updating code here accordingly.

### Why?
So project compiles :-)

----


OliverPearmain-Pusher:
Sorry, code was sound, I just forgot to `carthage update` 😞.  Should be good now.

PR #261, opened 2020-01-31
Replace Carthage with SPM

grzesiekko:
### What?

- Replace Carthage dependencies with Swift Package Manager Dependencies
- Replace Mockingjay with OHHTTPStubs

### Why?

Stability of our unit tests and build process.

----


OliverPearmain-Pusher:
> Replace Mockingjay with OHHTTPStubs

Nice one 🤛

Cartfile
```diff
@@ -1,2 +1,2 @@
 github "pusher/pusher-platform-swift" "v1"
-github "pusher/push-notifications-swift" ~> 3.0.1
+github "pusher/push-notifications-swift" ~> 3.0.4
```
OliverPearmain-Pusher: Don't we want to remove the Cartfile altogether if `pusher-platform-swift` and `pusher-notifications-swift` are now coming from SPM?

Chatkit.xcodeproj/project.pbxproj
```diff
@@ -2027,6 +1935,47 @@
...
+				branch = v1;
+				kind = branch;
+			};
+		};
+		22B2D2BF23E3004B0085D79C /* XCRemoteSwiftPackageReference "OHHTTPStubs" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/AliSoftware/OHHTTPStubs.git";
+			requirement = {
+				kind = upToNextMajorVersion;
+				minimumVersion = 9.0.0;
```
OliverPearmain-Pusher: Should `pusher-nofications-swift` be in here?  Perhaps we dropped it or perhaps its not SPM compatible?

Chatkit.xcodeproj/project.pbxproj
```diff
@@ -1280,9 +1254,11 @@
 			buildRules = (
 			);
 			dependencies = (
-				73A9391823E0A7440025AF90 /* PBXTargetDependency */,
 			);
 			name = "Test Utilities";
+			packageProductDependencies = (
+				22B2D2BC23E2FEE00085D79C /* PusherChatkit */,
```
OliverPearmain-Pusher: Is there a reason that you’ve linked the `PusherChatkit` *package* to `Test Utililties` as opposed to just adding the `Chatkit` target as a dependency?  (It seems a little circular to have the final package linked back into one of the projects own targets)

Screenshot 2020-01-31 at 12 34 26

vs

Screenshot 2020-01-31 at 12 34 38

NOTE using the `dependency` option does require chaning build setting to `SWIFT_INSTALL_OBJC_HEADER` to `NO` to get things to compile.  Howeve since we aren't supporting objective-c I don't think this is an issue?

Chatkit.xcodeproj/project.pbxproj
```diff
@@ -722,6 +695,13 @@
 			path = Chatkit;
 			sourceTree = "<group>";
 		};
+		22B2D2AE23E2FBF10085D79C /* Frameworks */ = {
+			isa = PBXGroup;
+			children = (
+			);
+			name = Frameworks;
+			sourceTree = "<group>";
+		};
```
OliverPearmain-Pusher: Really not a big deal but we this added `Frameworks` group is redundant.

Cartfile
grzesiekko: This file was used for 2 different reasons:
 - To define public dependencies of our SDK for everyone integrating using Carthage. This is still required.
 - To provide us with `PusherPlatform` and `PushNotifications` frameworks when doing the development of `PusherChatkit`. This is no longer required because Xcode fetches dependencies using SPM.

Chatkit.xcodeproj/project.pbxproj
grzesiekko: `PushNotifications` is compatible with SPM. I assumed that we will add this package when we will be implementing push notifications, but I can add now if you want.

`PushNotifications` is already a dependency defined by `Package.swift`.

Chatkit.xcodeproj/project.pbxproj
grzesiekko: This has been added automatically by Xcode. I will remove this group.

Chatkit.xcodeproj/project.pbxproj
grzesiekko: It was because of `SWIFT_INSTALL_OBJC_HEADER` being set to `YES`. I will change that, but `Chatkit` should also be included in `Link Binary With Libraries` section, because `Test Utilities` code actually links to this framework.

Chatkit.xcodeproj/project.pbxproj
OliverPearmain-Pusher: 👍 lets not add it until we need it

Cartfile
OliverPearmain-Pusher: > To define public dependencies of our SDK for everyone integrating using Carthage. This is still required.
Aha, cool, wasn't aware of this one. Thanks for educating me.

Chatkit.xcodeproj/project.pbxproj
OliverPearmain-Pusher: Thanks.  It still compiles without being linked but best to explicitly link it anyway 👍

PR #264, opened 2020-02-05
Added `Functional Tests` target 

OliverPearmain-Pusher:
### What?
- Added `Functional Tests` target (albeit empty for now) in
- Added new `ALL Tests` scheme

### Why?
- `Functional Tests` target is added in preparatiom of some forthcoming functional tests.
- `ALL Tests` scheme allows to execute all tests across the project

----


Chatkit/Networking/Model/Subscription/Wire.Event.EventType.swift
```diff
@@ -20,10 +20,6 @@ extension Wire.Event {
         
         // Presence Subscription
         case presenceState(event: Wire.Event.PresenceState)
-        
-        var description: String {
-            return String(describing: self)
-        }
```
OliverPearmain-Pusher: I removed this because it was redundant and affecting covergage.

System Tests/Tests/Authentication/Token Providers/TestTokenProviderTests.swift
```diff
@@ -39,7 +40,7 @@ class TestTokenProviderTests: XCTestCase {
             return
         }
         
-        XCTAssertTrue(token.token.count > 0)
+        XCTAssertTrue(token.value.count > 0)
```
OliverPearmain-Pusher: Compilation fix (that hadn't been spotted before).

Chatkit.xcodeproj/xcshareddata/xcschemes/ALL Tests.xcscheme
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
grzesiekko: We could potentially add all 3 test targets to `Chatkit` run scheme and have individual run schemes for each test target. An open question is whether we should run `System Tests` on our CI for every build or should this be triggered manually (or scheduled)? What do you think?

Chatkit.xcodeproj/project.pbxproj
```diff
@@ -407,6 +435,16 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		737E3F8023EAC9E000B50837 /* Frameworks */ = {
+			isa = PBXFrameworksBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				737E3F9523EACACB00B50837 /* PusherChatkit.framework in Frameworks */,
+				737E3F9423EACACB00B50837 /* Environment.framework in Frameworks */,
```
grzesiekko: Why do we need `Environment` in `Functional Tests`?

OliverPearmain-Pusher: Good question.  Its not required, I've removed it. (I had copied the `Unit Tests` setup. The `Unit Tests` needs it because it actually performs some tests on the content of `Environment`.  `Functional Tests` do not.)

Chatkit.xcodeproj/xcshareddata/xcschemes/ALL Tests.xcscheme
OliverPearmain-Pusher: Yes that is an option.  I'm not sure I have formed a solid opinion on whether System tests should run on CI.   I think it might be good to run them but then we'd need to be concerned with the setup of the Instance they run against and also the possibility of flakey tests.  

At the very least we should *compile* the System tests on CI to ensure any refactoring hasn't broken compilation.

OliverPearmain-Pusher: OK, finally got the "Swift Environment Variables Injector" working.  I have updated the `Chatkit` scheme to run all the tests and removed the `ALL Tests` scheme.  Builds are now passing on Bitrise.

PR #266, opened 2020-02-07
Joined rooms reducers

grzesiekko:
### What?

- Add reducers that convert `initialState`, `addedToRoom`, `removedFromRoom`, `readStateUpdated`, `roomDeleted` and `RoomUpdated` events into an internal state.
- Remove references to tests that no longer exist in the codebase.

### Why?

We need to transform our wire models into an internal state.

----


Chatkit/Dependencies/Dependencies.swift
```diff
@@ -86,4 +87,9 @@ class ConcreteDependencies: Dependencies {
     var store: Store {
         return dependencyFactory.resolve(Store.self, dependencies: self)
     }
+    
+    var reductionManager: ReductionManager {
+        return dependencyFactory.resolve(ReductionManager.self, dependencies: self)
```
OliverPearmain-Pusher: You need to update the tests in`ConcreteDependencyTests` to account for this new property.  And I think those tests might fail because you've not `registered` the `ConcreteReductionManager` in the initialiser. 

Also whilst you're in `ConcreteDependencyTests` would you mind correcting some mistakes I've made 🙏🏼...
- the file/class should be name `ConcreteDependenciesTests` 
- it should live in the `Tests/Dependencies` directory/group but its currently in `Tests/Redux/Store`

grzesiekko: Sure, I will have a look into that.

Chatkit/Redux/State/RoomState.swift
```diff
@@ -0,0 +1,19 @@
+
+struct RoomState: State {
+    
+    // MARK: - Properties
+    
+    let identifier: String
+    let name: String
+    let isPrivate: Bool
+    let pushNotificationTitle: String?
```
OliverPearmain-Pusher: Are you missing `createdById`?

grzesiekko: I thought that we want to drop `createdBy` from our API. Is this still correct @mdpye?

Chatkit/Redux/State/UserState.swift
```diff
@@ -0,0 +1,15 @@
+
+enum UserState {
+    
+    case empty
```
OliverPearmain-Pusher: I not convinced that `empty` belongs here otherwise when we implement the list of users for a room it will be possible to have `empty` users in that list, which makes no sense.

We obviously need something to model the "no current user" state.  Would it be better to drop `case empty` here and just make `ChatState.currentUser` an optional?

grzesiekko: Sure, that's an option. I am also not convinced about naming the two cases below as `populated` and `partial`.

Chatkit/Redux/State/ChatState.swift
```diff
@@ -0,0 +1,17 @@
+
+struct ChatState: State {
+    
+    // MARK: - Properties
+    
+    let currentUser: UserState
+    let joinedRooms: RoomListState
```
OliverPearmain-Pusher: I think there will eventually be other `Room`s in the store besides just `joinedRoom` (e.g. `joinableRooms`).  So I think we may need to model this differently?  

Perhaps there could be a master list of rooms and then other things key into that list by identifier, for example... 
```
let rooms: RoomListState
let joinedRoomIds: [String]
let joinableRoomIds: [String]
```

OliverPearmain-Pusher: The same probably applies to `currentUser`

```
let users: [UserState]
let currentUserId: String?
```

grzesiekko:
Do you guys think that reducers could potentially benefit from introducing middleware (one of Redux concepts)? With middleware we would be able to prepare/parse/etc. our actions before passing them to reducers. This could simplify reducers and the whole transformation process from wire models to the internal state.

Chatkit/Redux/State/RoomListState.swift
```diff
@@ -0,0 +1,14 @@
+
+struct RoomListState: State {
```
OliverPearmain-Pusher: Is there a reason I'm missing why this `RoomListState` is necessary rather than working with just a `[RoomState]` directly?

Chatkit/Redux/State/ChatState.swift
grzesiekko: I don't remember what have we decided about joinable rooms. I do remember, however, that we don't want to persist joinable rooms anywhere or reuse that list. I am not sure how that relates to the internal state.

Chatkit/Redux/State/RoomListState.swift
grzesiekko: That's required, because reducers are not fit for mapping between wire model and internal state. In redux this should be done using middleware, but we don't have such layer.

grzesiekko: We could potentially work around this issue, but this would probably involve introducing a concept of sub-actions. What are your thoughts on this problem?

OliverPearmain-Pusher: I don't think I fully understand the points you're making, very sorry 🥺.  Perhaps I need to read up on middleware.

I had a go and was able to get rid of the `RoomListState` entity without much effort.  See here...
https://github.com/pusher/chatkit-swift/commit/8e1ab00594e36ab9274151d1bf99de850bb46888

Perhaps you are suggesting that we will require the `RoomListState` so that **in future** we have the option of using "middleware"?

grzesiekko: I agree that this is possible to code, but the change you suggested breaks the definition of a reducer (and perhaps that's ok, because we don't want to follow redux?). 

Reducer should take an action and a state and return a new/unmodified state. In order to get rid of `RoomListState`, which annoys me as well, we would need to pass to the reducer a list of states (like you suggested) or break the action into sub-actions. Either of these approaches feels like a deviation from the classical redux. Are we ok with that?

OliverPearmain-Pusher: I not sure how its break the definition?  `[RoomState]` is itself a `State` as defined here...

https://github.com/pusher/chatkit-swift/commit/8e1ab00594e36ab9274151d1bf99de850bb46888#diff-a2e538b46ffc792fe7c83fd97e1d1284R22

Chatkit/Redux/State/UserState.swift
OliverPearmain-Pusher: Not sure I have a strong opinion on the `populated` and `partial` bit at the moment.  I'd be happy for that to stay as it in this PR and we can address later when we are coding the actions/reducers that work with partial models.

Chatkit/Redux/State/ChatState.swift
OliverPearmain-Pusher: If that is the case then `let joinedRooms: RoomListState` is all good.  Any thoughts on the users though?

Test Utilities/Doubles/Reducer.swift
```diff
@@ -0,0 +1,48 @@
...
+    private var reducer_expectedState: StateType
+    public private(set) var reducer_actionLastReceived: ActionType?
+    public private(set) var reducer_stateLastReceived: StateType?
+    public private(set) var reducer_actualCallCount: UInt = 0
+    
+    public init(reducer_expectedState: StateType,
+                reducer_expectedCallCount: UInt = 0,
+                file: StaticString = #file, line: UInt = #line) {
+        
+        self.reducer_expectedState = reducer_expectedState
```
OliverPearmain-Pusher: `reducer_expectedState` is never used and can be removed.

OliverPearmain-Pusher: Sorry, I am wrong, it is used.  I think it was the variable name that threw me as I thought you were checking this value against the _incoming_ state.  

I have been using `<func_name>_<object>ToReturn` in other Stubs, so for consistency perhaps you might rename to `reducer_stateToReturn`?

Unit Tests/Tests/Redux/Reducers/ConcreteReductionManagerTests.swift
```diff
@@ -0,0 +1,207 @@
...
+                readStates: [],
+                memberships: []
+            )
+        )
+        
+        /******************/
+        /*----- WHEN -----*/
+        /******************/
+        
+        let result = sut.reduce(action: action, state: currentState)
```
OliverPearmain-Pusher: Is `newState` or `updatedState` a better variable name than `result`?  (this applies to all your tests).

Chatkit/Redux/State/RoomState.swift
mdpye: Yeah, let's try. I think it's listed in the discovery doc

mdpye:
> Do you guys think that reducers could potentially benefit from introducing middleware (one of Redux concepts)? With middleware we would be able to prepare/parse/etc. our actions before passing them to reducers. This could simplify reducers and the whole transformation process from wire models to the internal state.

I was leary of the idea of using middlewares to fire network calls, but as pure mappers, they seem to fit the bill, and the architecture.

Unit Tests/Tests/Redux/Reducers/ConcreteReductionManagerTests.swift
OliverPearmain-Pusher: Actually I've changed my mind 🤣 .  I think `inputState` and `outputState` might be nicer.   So in the of this case of this test you'd rename as follows:

`expectedState` -> `initialStateReducer_inputState`
`currentState` -> `inputState`
`result` -> `outputState`

and probably add `let expectedState = initialStateReducer_inputState` at the top of the "THEN" section.

Chatkit/Redux/Actions/ReceivedRemovedFromRoomAction.swift
```diff
@@ -0,0 +1,12 @@
+
+struct ReceivedRemovedFromRoomAction: Action {
+    
+    // MARK: - Properties
+    
+    let event: Wire.Event.RemovedFromRoom
```
OliverPearmain-Pusher: Do we ever foresee these concrete `Action`s having more than just an `event` property?  Only I'm wondering if we can get rid of all the boilerplate of these "wrapper" struct's and just make the event itself conform to `Action`?

`extension Wire.Event.RemovedFromRoom: Action {}`

Chatkit/Redux/State/RoomState.swift
OliverPearmain-Pusher: 👍  I'll make a note to drop it form the Wire models also.

Chatkit/Dependencies/Dependencies.swift
grzesiekko: Done

Chatkit/Redux/State/UserState.swift
grzesiekko: I replaced `empty` with an optional value.

Chatkit/Redux/State/ChatState.swift
grzesiekko: I have added the list of all users to `ChatState`.

Chatkit/Redux/State/RoomListState.swift
grzesiekko: `extension Array: State where Element == RoomState {}` is a less implicit version of `RoomListState`.

I have replaced `RoomListState` with the above extension.

Test Utilities/Doubles/Reducer.swift
grzesiekko: Done

Unit Tests/Tests/Redux/Reducers/ConcreteReductionManagerTests.swift
grzesiekko: Done

Chatkit/Redux/Actions/ReceivedRemovedFromRoomAction.swift
grzesiekko: We might have other actions in future that will not have subscription events attached to them.

OliverPearmain-Pusher:
This is a really solid implementation thats well tested and completely achieves the objective. :thumbsup: 👏 

However I do have a reservation with the approach.  It appears the `Reducer.ReducerType` typealias is being used as a blueprint for the Reducers but the reality is that this is only the case **at the first layer**.   In subsequent/lower layers:

- Its possible to violate the requirement of taking an `Action` and `State`.
- Its not possible to inject child reduce functions (therefore they must be called directly & we cannot unit test intermediate level reducers individually)

My concern is that this will impact our ability to scale this in future.

I appreciate this is a tricky problem to solve so rather than be that guy that just criticises without suggesting an alternative I spent some further time on this over the weekend.  I think I have come up with something which could solve these problems (and has other benefits).

You already have `static func` implementations as extensions.  These can stay but in addition to this…
- we update the `static func`s to also take a `dependencies` parameter (on which child reduce func’s and middleware can live)
- every reducer func gets an associated `HasReducer_X` protocol (so that it can be stubbed in tests)
- we have a few ways we can define the “Concrete” implementations of the dependencies.  (I’m using `ReducerDependencies` below)

```
protocol HasReducer_UserSubscription_InitialState {

    var reducer_userSubscription_initialState: (
        ReceivedInitialStateAction,
        ChatState,
        HasReducer_Model_User_forInitialState & HasReducer_Model_Rooms_forInitialState
    ) -> ChatState { get }
}

extension Reducer.UserSubscription {

    static func initialState(
        action: ReceivedInitialStateAction,
        state: ChatState,
        dependencies: HasReducer_Model_User_forInitialState & HasReducer_Model_Rooms_forInitialState
    ) -> ChatState {

        let currentUser = dependencies.reducer_model_user_forInitialState(action, state.currentUser)
        let joinedRooms = dependencies.reducer_model_rooms_forInitialState(action, state.joinedRooms)

        return ChatState(currentUser: currentUser, joinedRooms: joinedRooms)
    }
}
```

```
protocol ReducerDependencies:
    HasReducer_Master &
    HasReducer_Model_User_forInitialState &
    HasReducer_Model_Rooms_forInitialState &
    HasReducer_Model_Rooms_forRemovedFromRoom &
    HasReducer_UserSubscription_InitialState &
    HasReducer_UserSubscription_RemovedFromRoom
{}

struct ConcreteReducerDependencies: ReducerDependencies {
    let reducer_master = Reducer.master
    let reducer_model_user_forInitialState = Reducer.Model.user
    let reducer_model_rooms_forInitialState = Reducer.Model.rooms_forInitialState
    let reducer_model_rooms_forRemovedFromRoom = Reducer.Model.rooms_forRemovedFromRoom
    let reducer_userSubscription_initialState = Reducer.UserSubscription.initialState
    let reducer_userSubscription_removedFromRoom = Reducer.UserSubscription.removedFromRoom
}

```
This approach also has the benefit of solving these other general issues…
- Provides a way to inject middleware/parsers (via `dependencies`) which has been mentioned in this PR already
- The top level reducer can actually become a true reducer (no need for a `ReductionManager`)

There is certainly more boilerplate with this approach but I think that might be worth it for the problems it solves.  We _might_ be able to reduce (excuse the pun) some of this with some `typealiases`

What do you think of this approach?

PS: 
- I have not attempted to write any unit tests against this approach yet.  So I am assuming this will work with stubs and tests. 🙏 
- I can put together a branch of this if you want to see more 
 - The top-level reducer will look like the below (to demonstrate it too is a reducer)...

```
protocol HasReducer_Master {
    
    var reducer_master: (
        Action,
        ChatState,
        ReducerDependencies
    ) -> ChatState { get }
}

extension Reducer {
    
    static func master(
        action: Action,
        state: ChatState,
        dependencies: ReducerDependencies
    ) -> ChatState {
        
        if let action = action as? ReceivedInitialStateAction {
            return dependencies.reducer_userSubscription_initialState(action, state, dependencies)
        }
        else if let action = action as? ReceivedRemovedFromRoomAction {
            return dependencies.reducer_userSubscription_removedFromRoom(action, state, dependencies)
        }
        
        return state
    }
}
```

grzesiekko:
@OliverPearmain-Pusher I will try to add DI to reducers. If you have something ready on some branch then please push the code to the remote repository.

Chatkit/Dependencies/Dependencies.swift
```diff
@@ -3,9 +3,20 @@ import struct PusherPlatform.PPSDKInfo
...
+    HasRoomListReducer &
+    HasUserSubscriptionInitialStateReducer &
+    HasUserSubscriptionAddedToRoomReducer &
+    HasUserSubscriptionRemovedFromRoomReducer &
+    HasUserSubscriptionRoomUpdatedReducer &
+    HasUserSubscriptionRoomDeletedReducer &
+    HasUserSubscriptionReadStateUpdatedReducer
 {}
 
+typealias NoDependencies = Any
```
OliverPearmain-Pusher: Nice 👍

Chatkit/Redux/Reducers/MasterReducer.swift
```diff
@@ -0,0 +1,56 @@
...
+            return state
+        }
+        
+    }
+    
+}
+
+// MARK: - Dependencies
+
+protocol HasMasterReducer {
```
OliverPearmain-Pusher: I've noticed that you've moved all the `protocol Has*` to the bottom of the file where-as I would expect to see relevant protocols for a given class at the top.  (for example you'd normally define `Delegate` protocols above the class itself).

This also keeps things consistent with the pre-existing `InstanceLocator`, `Store` and `StoreBroadcaster`.

Chatkit/Redux/Actions/Action.swift
```diff
@@ -0,0 +1,2 @@
+
+class Action {}
```
OliverPearmain-Pusher: Is the reason `Action` has become a class because it wasn't possible to specify a `typealias ActionType = Action` when using a protocol?

OliverPearmain-Pusher: Because there is **a lot** baggage that comes with this alteration.

Test Utilities/Doubles/State.swift
```diff
@@ -0,0 +1,23 @@
+@testable import PusherChatkit
+
+public struct TestState {
```
OliverPearmain-Pusher: This shouldn't live in the `Doubles` group/directory because it isn't a test double, this is test data itself.  

Its generally bad practice to put test data into a common place because it "hides" the data away from the test.  Developers should be able to read the test data at the test site to determine whats going on.  

Having looked a little closer at the use of these values I've determined that its used almost exclusively to fill in blanks that aren't important for a given test. 

Therefore you can simply remove this class/file if you: 

- replace all use of `TestState.user`  with `.empty`
- replace all use of  `TestState.userIdentifier` with `"alice"`
- replace all use of  `TestState.userList` with `.empty` except for `test_dispatch_withActionThatDoesChangeInternalState_delegateTriggered` where you'll need to use the value directly.

I've actually done all this for you on this branch...
https://github.com/pusher/chatkit-swift/tree/OKP/Reducers-remove-TestState

Test Utilities/Doubles/Reducer.swift
```diff
@@ -0,0 +1,58 @@
...
+        
+        super.init(file: file, line: line)
+    }
+    
+    public func reduce(action: T.ActionType, state: T.StateType, dependencies: T.DependenciesType) -> T.StateType {
+        reduce_actionLastReceived = action
+        reduce_stateLastReceived = state
+        reduce_actualCallCount += 1
+        
+        guard reduce_expectedCallCount > 0 else {
```
OliverPearmain-Pusher: Really no big deal but I've found myself using this...
`guard reduce_actualCallCount <= reduce_expectedCallCount else {`
... because it allows you make `reduce_expectedCallCount` a `let`

Test Utilities/Doubles/Action.swift
```diff
@@ -0,0 +1,9 @@
+@testable import PusherChatkit
+
+public class DummyAction: Action {
```
OliverPearmain-Pusher: I'd probably name this `FakeAction` (its not inheriting from `DummyBase` and doesn't cause any issues if interacted with).

Chatkit/Redux/State/UserState.swift
OliverPearmain-Pusher: For reference, `.empty` is now back as per discussion here:
https://pusher.slack.com/archives/CSVSXJNEM/p1581509086001500

Chatkit/Redux/State/UserListState.swift
```diff
@@ -0,0 +1,14 @@
+
+struct UserListState: State {
+    
+    // MARK: - Properties
+    
+    let users: [String : UserState]
+    
+    static let empty: UserListState = UserListState(users: [:])
+
```
OliverPearmain-Pusher: Any merit in adding a convience constructor that takes an **array** of users and maps the keys from the user identifier?

OliverPearmain-Pusher: Same applies to `RoomListState`

Chatkit/Redux/Reducers/Model/RoomListReducer.swift
```diff
@@ -0,0 +1,145 @@
...
+        
+        typealias ActionType = Action
+        typealias StateType = RoomListState
+        typealias DependenciesType = NoDependencies
+        
+        // MARK: - Reducer
+        
+        static func reduce(action: ActionType, state: StateType, dependencies: DependenciesType) -> StateType {
+            if let action = action as? InitialStateAction {
+                return self.reduce(action: action, state: state, dependencies: dependencies)
```
OliverPearmain-Pusher: `self.` is redundant here (and other calls to `reduce` below)

Chatkit/Redux/Reducers/Model/RoomListReducer.swift
```diff
@@ -0,0 +1,145 @@
...
+                $0[$1.identifier] = RoomState(identifier: $1.identifier,
+                                              name: $1.name,
+                                              isPrivate: $1.isPrivate,
+                                              pushNotificationTitle: $1.pushNotificationTitleOverride,
+                                              customData: $1.customData,
+                                              lastMessageAt: $1.lastMessageAt,
+                                              readSummary: readSummary,
+                                              createdAt: $1.createdAt,
+                                              updatedAt: $1.updatedAt)
+            }
```
OliverPearmain-Pusher: This would be a lot more readable with named args....
```
let readSummaries = action.event.readStates.reduce(into: [String : ReadSummaryState]()) { dict, readSummaryState in
    dict[readSummaryState.roomIdentifier] = ReadSummaryState(unreadCount: readSummaryState.unreadCount)
}

let rooms = action.event.rooms.reduce(into: [String : RoomState]()) { dict, roomState in
    let readSummary = readSummaries[roomState.identifier] ?? .empty
    
    dict[roomState.identifier] = RoomState(identifier: roomState.identifier,
                                           name: roomState.name,
                                           isPrivate: roomState.isPrivate,
                                           pushNotificationTitle: roomState.pushNotificationTitleOverride,
                                           customData: roomState.customData,
                                           lastMessageAt: roomState.lastMessageAt,
                                           readSummary: readSummary,
                                           createdAt: roomState.createdAt,
                                           updatedAt: roomState.updatedAt)
}
```

Unit Tests/Tests/Redux/Reducers/Model/UserReducerTests.swift
```diff
@@ -0,0 +1,79 @@
+import XCTest
+import TestUtilities
+@testable import PusherChatkit
+
+class UserReducerTests: XCTestCase {
+    
+    // MARK: - Tests
+    
+    func test_reduce_withCurrentStateAndInitialStateAction_returnsModifiedState() {
```
OliverPearmain-Pusher: A lot of your tests include `withCurrentStateAnd` in their name which I don't think adds any value (because there has to be a current state thats passed to every `redude` function.  Perhaps you drop this to make the method names more succinct (like in `MasterReducerTests`)

Unit Tests/Tests/Redux/Reducers/User Subscription/AddedToRoomReducerTests.swift
```diff
@@ -0,0 +1,176 @@
...
+                        createdAt: .distantPast,
+                        updatedAt: .distantPast
+                    )
+                ]
+            ),
+            users: TestState.userList
+        )
+        
+        XCTAssertEqual(outputState, expectedState)
+        XCTAssertEqual(stubReducer.reduce_actualCallCount, 1)
```
OliverPearmain-Pusher: **Some general points on all your reducer tests:**

- keep your inputs and outputs as minimal as possible. I've noticed some of the tests are unnecessarily verbose.  Example below. 

- When defining `expectedState`, rather then rewriting out raw values, if you're expecting a certain State property to be...
   - unchanged, then use `inputState.property` for the value
   - changed, then you can probably use `reduce_stateToReturn` for the value

- the only thing that needs ascerting in the `UserSubscription` reducer tests is that the sub-reducers are invoked with the appropriate arguments.  So 

   - Its not essential to emulate what would happen in the real world with your input and output data.  Just ensure that things that might change do actually change in the output
   - Don't forget to make these assertions:

```
	XCTAssertEqual(stubReducer.reduce_stateLastReceived, inputState.joinedRooms)
	XCTAssertTrue(stubReducer.reduce_actionLastReceived === action)  // <-- this is ugly because of `class Action`
```

--------------------------

So to take this specific `test_reduce_withCurrentStateAndAddedToRoomAction_returnsStateFromDedicatedReducer` test as an example.  

I have been able to improve the assertions and reduce the lines of code in this single test from 160+ to 75 as follows:

```
        /******************/
        /*---- GIVEN -----*/
        /******************/
        
        let inputState = MasterState(
            currentUser: .empty,
            joinedRooms: .empty,
            users: .empty
        )
        
        let action = AddedToRoomAction(
            event: Wire.Event.AddedToRoom(
                room: Wire.Room(
                    identifier: "new-room",
                    name: "New",
                    createdById: "random-user",
                    isPrivate: false,
                    pushNotificationTitleOverride: nil,
                    customData: nil,
                    lastMessageAt: .distantPast,
                    createdAt: .distantPast,
                    updatedAt: .distantPast,
                    deletedAt: .distantPast),
                membership: Wire.Membership(
                    roomIdentifier: "new-room",
                    userIdentifiers: []
                ),
                readState: Wire.ReadState(
                    roomIdentifier: "new-room",
                    unreadCount: 30,
                    cursor: nil
                )
            )
        )
        
        let reducer_stateToReturn = RoomListState(
            rooms: [
                "new-room" : RoomState(
                    identifier: "new-room",
                    name: "New",
                    isPrivate: false,
                    pushNotificationTitle: nil,
                    customData: nil,
                    lastMessageAt: .distantPast,
                    readSummary: .empty,
                    createdAt: .distantPast,
                    updatedAt: .distantPast
                )
            ]
        )
        
        let stubReducer = StubReducer<Reducer.Model.RoomList>(reduce_stateToReturn: reducer_stateToReturn, reduce_expectedCallCount: 1)
        
        let dependencies = DependenciesDoubles(roomListReducer: stubReducer.reduce)
        
        /******************/
        /*----- WHEN -----*/
        /******************/
        
        let outputState = Reducer.UserSubscription.AddedToRoom.reduce(action: action, state: inputState, dependencies: dependencies)
        
        /******************/
        /*----- THEN -----*/
        /******************/
        
        let expectedState = MasterState(
            currentUser: inputState.currentUser,
            joinedRooms: reducer_stateToReturn,
            users: inputState.users
        )
        
        XCTAssertEqual(outputState, expectedState)
        XCTAssertEqual(stubReducer.reduce_actualCallCount, 1)
        XCTAssertEqual(stubReducer.reduce_stateLastReceived, inputState.joinedRooms)
        XCTAssertTrue(stubReducer.reduce_actionLastReceived === action)
```

Chatkit/Redux/Actions/Action.swift
OliverPearmain-Pusher: This approach is my preference (see last commit on this branch)
https://github.com/pusher/chatkit-swift/tree/OKP/Reducers-Action-not-class

Why/How: 
- `Action` is still a protocol, we've added a `MasterAction` enum
- No reference types floating around
- No need to define custom `Equatable` implementations for every `Action`
- `MasterAction` enum prevents necessity to cast anything
- Its now possible to perform `XCTAssertEqual(masterAction, anotherMasterAction)`

OliverPearmain-Pusher:
OK done with review for now.  Going to get back to working on Subscriptions.

Chatkit/Redux/Reducers/MasterReducer.swift
grzesiekko: This is a matter of personal preference. One could argue that when a file is named `Name.swift` than the most important element of this file (the class/struct/enum/etc. `Name`) should be placed on top of the file. Such approach also allows pragma marks to be places in a way which then nicely integrates with Xcode's navigation.

Chatkit/Redux/Actions/Action.swift
grzesiekko: Yes, this is due to the mentioned `typealias`.

`MasterAction` has one big downside which I would prefer to avoid. It does not allow to constrain input parameters to a specific subtype of `Action`.

Test Utilities/Doubles/State.swift
grzesiekko: In one of my previous PRs you asked me to centralize test data and now I am supposed to do the opposite. Sure...why not...

Test Utilities/Doubles/Action.swift
grzesiekko: I will rename this.

Test Utilities/Doubles/Reducer.swift
grzesiekko: This has been copied from other classes that are currently in our codebase.

Chatkit/Redux/State/UserListState.swift
grzesiekko: This is a nice to have, so let's add this in future.

Chatkit/Redux/Reducers/Model/RoomListReducer.swift
grzesiekko: The usage of an implicit/explicit `self` is a matter o personal preference.

Chatkit/Redux/Reducers/Model/RoomListReducer.swift
grzesiekko: This is a matter o personal preference.

Unit Tests/Tests/Redux/Reducers/Model/UserReducerTests.swift
grzesiekko: Sure, I can drop this.

Test Utilities/Doubles/State.swift
OliverPearmain-Pusher: I don't recall making such a request.  Are you are to direct me to it please?

Chatkit/Redux/Actions/Action.swift
OliverPearmain-Pusher: I think it does.  If you check my branch you'll see that `UserSubscription` reducers all use constrained `Action`s, for example...
```
typealias ActionType = ReadStateUpdatedAction 
```
https://github.com/pusher/chatkit-swift/blob/8127836165fd9f06b8d2d445675ee9b5c90ecf2f/Chatkit/Redux/Reducers/User%20Subscription/ReadStateUpdatedReducer.swift#L8

`MasterReducer` and `Model` reducers use `MasterAction`.

Test Utilities/Doubles/Reducer.swift
OliverPearmain-Pusher: Yes its probably copied from code I wrote.  I didn't word it very well but what I was trying to say is that I've since evolved that implemented to avoid the necessity of a `var`.  If you don't want to change then its no bother.

Chatkit/Redux/Reducers/Model/RoomListReducer.swift
OliverPearmain-Pusher: OK cool, I have no issue if you like to use `self.`.  Just wasn't sure if it was intentional or not.

Chatkit/Redux/Reducers/MasterReducer.swift
OliverPearmain-Pusher: Sure, everything can be attributed to personal preference.  I'm happy to be flexible on these minor details but I do like things to be consistent.  So I think we either need to update all the existing `protocol Has*` moving them to the bottom or move these back to the top. 

Not terribly fussed either way but last time we didn't share opinion on a given personal preference the decision was made not to change things for change sake and to continue with the approach that had set a precedent. 

I'll leave it up to you.  I'm happy to approve either way.

Test Utilities/Doubles/State.swift
grzesiekko: > Do you think we should centralised the value "joe"? Happy for this to be done in another PR though.

> I think you added TestUser.joe but you forgot to consume it here :-)

https://github.com/pusher/chatkit-swift/pull/246

Chatkit/Redux/Actions/Action.swift
grzesiekko: This approach means that we introduce subactions.

Test Utilities/Doubles/State.swift
OliverPearmain-Pusher: Just revisiting this after my holidays.  Regarding the quoted comment (https://github.com/pusher/chatkit-swift/pull/246#discussion_r370168330), I think you have misinterpreted or taken things out of context.  The comment relates to a value in `Demo/Controllers/RoomListViewController.swift`, a UIViewController of the Demo app, and therefore is of no relevance to **test** data.

OliverPearmain-Pusher:
There were a number of comments unaddressed & unanswered when this was merged.  Was a decision made to disregard these or are we planning to address them in future?

PR #271, opened 2020-03-05
OKP SubscriptionStateUpdatedAction

OliverPearmain-Pusher:
### What?

Updated `Subscription` so that it notifies the store when its State changes 

### Why?

So it ultimately updates the ConnectionState of the AuxiliaryState and can be reported to Repositories/ViewModels.


grzesiekko:
I think that it would be good to merge our integration branch into this PR to make things up-to-date.

OliverPearmain-Pusher:
This should now be up-to-date with the integration branch and is ready for a proper review please.

Chatkit/Connectivity Monitor/ConnectivityMonitor.swift
```diff
@@ -18,20 +17,23 @@ class ConcreteConnectivityMonitor: ConnectivityMonitor {
...
-    private(set) var connectionState: ConnectionState
+    
+    private var connectionState: ConnectionState!
     
     private let dependencies: Dependencies
     
     weak var delegate: ConnectivityMonitorDelegate?
     
     // MARK: - Initializers
+    static func makeWithInitialValue(subscriptionType: SubscriptionType, dependencies: Dependencies) -> (ConcreteConnectivityMonitor, ConnectionState) {
```
grzesiekko: This comment is probably unimportant as `ConnectivityMonitor` will be soon replaced by something else, but I think it would be go to either `typealias` the tuple or replace the tuple with a dedicated `struct`.

OliverPearmain-Pusher: I am quite happy to update to a typealias, something like... 

```
    typealias ConnectivityMonitorAndInitialConnectionState = (
        connectivityMonitor: ConcreteConnectivityMonitor,
        initialConnectionState: ConnectionState
    )
```
```
static func makeWithInitialValue(subscriptionType: SubscriptionType, 
          dependencies: Dependencies) -> ConnectivityMonitorAndInitialConnectionState
```

... however because I don't fully understand the motiviation I'm not sure how you prefer/imagine the client code to look.  

I have the option of leaving things unchanged....
```
let (connectivityMonitor, initialConnectionState) = ConcreteConnectivityMonitor.makeWithInitialValue(subscriptionType: .user, dependencies: self.dependencies)

return ConcreteJoinedRoomsRepository(buffer: buffer,
             connectivityMonitor: connectivityMonitor,
             connectionState: initialConnectionState,
             dependencies: self.dependencies)
```

or one could work with the typealias in a couple of different ways.  Like this...

```
let connectivityMonitorAndInitialConnectionState = ConcreteConnectivityMonitor.makeWithInitialValue(subscriptionType: .user, dependencies: self.dependencies)

return ConcreteJoinedRoomsRepository(buffer: buffer,
             connectivityMonitor: connectivityMonitorAndInitialConnectionState.connectivityMonitor,
             connectionState: connectivityMonitorAndInitialConnectionState.initialConnectionState
             dependencies: self.dependencies)
```

or this (updated the `ConcreteJoinedRoomsRepository` initializer also)...
```
let connectivityMonitorAndInitialConnectionState = ConcreteConnectivityMonitor.makeWithInitialValue(subscriptionType: .user, dependencies: self.dependencies)

return ConcreteJoinedRoomsRepository(buffer: buffer,
             connectivityMonitorAndInitialConnectionState: connectivityMonitorAndInitialConnectionState,
             dependencies: self.dependencies)
```

Any thoughts.

grzesiekko: Let's keep things as they are (option 1) and only add a `typealias`. It helps with readability and reusability of the type.

OliverPearmain-Pusher: Thanks, I just pushed that change.

PR #272, opened 2020-03-06
Joined rooms demo

grzesiekko:
### What?

Basic integration of `JoinedRoomsViewModel` with the demo.

### Why?

Roadmap

----


Demo/Controllers/LoginViewController.swift
```diff
@@ -14,13 +14,31 @@ class LoginViewController: UIViewController {
...
         super.viewDidAppear(animated)
         self.displayKeyboard()
     }
     
+    // MARK: - Navigation
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        super.prepare(for: segue, sender: sender)
+
+        if segue.identifier == LoginViewController.loginSegue {
```
OliverPearmain-Pusher: Its not technically a change introduced by this PR but do you think we could rename `loginSegue` to `loginSegueId` so its type/value is a little more explicit?

Demo/Controllers/LoginViewController.swift
```diff
@@ -14,13 +14,31 @@ class LoginViewController: UIViewController {
...
+            guard let roomListViewController = segue.destination as? RoomListViewController else {
```
OliverPearmain-Pusher: This code doesn't work at present because `segue.destination` is a `UINavigationController`.  This works instead...
```suggestion
        if segue.identifier == LoginViewController.loginSegue {
            guard let navigationController = segue.destination as? UINavigationController,
                let roomListViewController = navigationController.viewControllers[0] as? RoomListViewController else {
```

Demo/Controllers/LoginViewController.swift
grzesiekko: Sure, no problem.

Demo/Controllers/RoomListViewController.swift
```diff
@@ -3,124 +3,176 @@ import PusherChatkit
...
-                let messageViewController = segue.destination as? MessageViewController else {
-                    return
-            }
-            
-            messageViewController.room = room
-            messageViewController.chatkit = self.chatkit
-        }
+    // MARK: - Private methods
+    
+    private func createViewModel() {
```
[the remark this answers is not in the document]
grzesiekko: I renamed the method.

